PR #181, opened 2014-08-01
cmake build (wip)

ksophocleous:
- cmake build file builds libsodium and tests (if requested) and also creates install targets
- removed manually-created visual studio solution and project files
- generate resource.rc on win32 platforms and build it into the DLL
- added ctest support
- added cpack support

Tests are build by default, to disable building tests use `cmake -DBUILD_LIBSODIUM_TESTS=off ..`. 

The simplified build output is forcing the build outputs to be in `bin` and `lib` on visual studio generators.

I do not have the chance to test it on OSX.

Tested with generators
- [x] Visual Studio 12 (win32)
- [x] Visual Studio 11 (win32)
- [x] Unix Makefiles (Cygwin)
- [x] MSYS (win32 not cygwin)
- [x] Unix Makefiles (Linux)
- [ ] MinGW (Linux cross compiling)


emmenlau:
Is the decision about the specific build system something that in time may be reconsidered? Please forgive if I'm being annoying with this kind request!

Its just my personal stance that I prefer generating build instructions myself, and I'm under the impression that currently that is not supported on MSVC platform for sodium? In recent years cmake has gained a lot of momentum and I'm just curious if there is a chance to promote cmake support?

Thanks a lot for your consideration!

PR #192, opened 2014-09-24
added gist examples into demos folder; you can figure out where this bel...

GraxRabble:
Here are the gists example. Might as well toss these in here. I noticed the documentation doesn't show how to use afternm and box.c does exactly that.


jedisct1:
Not forgetting about your pull request; sorry for the delay.
This will be added shortly.

PR #211, opened 2014-11-22
Automatic script to convert Libsodium to LibsodiumJS

CryptoNoob:
I hope this script help the people who want to convert the Libsodium to LibsodiumJS.
It needs more gateways between the C and the JS, but right now it is functional.

Keep Up the good work Frank! :D

### DIFF
--- a/dist-build/emscripten_tunned.sh
+++ b/dist-build/emscripten_tunned.sh
@@ -1,0 +1,120 @@
+#! /bin/sh
+
+
+#Preparing the enviroment
+echo "-----------------------------------------------------------------"
+echo "---> In order to test if you have all the packagues needed to compile, the script needs your user password:"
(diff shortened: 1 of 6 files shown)


jedisct1:
@BatikhSouri took a slightly different approach with a lot of inspiration from your code.
Let's try to combine the best of both worlds!

Batikh's use of JSON files to describe functions inputs and outputs is pretty neat.

A script that downloads and installs a full environment to compile and test the code is great, but it requires a very specific Linux distribution to start with. If we want to do this, I'd rather suggest using Docker or Vagrant.

Discussion on Batikh's work is here https://github.com/jedisct1/libsodium/pull/233 and discussions about the JS version should probably happen on Batikh's new repository or https://github.com/jedisct1/libsodium.js

Thanks!

PR #224, opened 2014-12-28
Add '-Wl,-pie' check to squash clang warnings

JackWink:
Clang on OSX doesn't like the `-pie` linker flag, instead it's expecting `-Wl,-pie`. This adds a check for `-Wl,-pie` linker flag support and opts to use it when valid.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -169,7 +169,6 @@ AS_IF([test "$enable_pie" != "no"],[
                 LDFLAGS="$LDFLAGS -pie"
             ])
         ])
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi Jack,

Sorry, but I had to revert this. It broke compilation on OpenBSD.

JackWink:
No problem, I'll check it out soon, any tips for getting an OpenBSD test environment set up? I don't mind burning up some AWS credit if that's an option/quicker than virtualbox/etc.

jedisct1:
VMware and Virtualbox work pretty well for this :)

Issue #178, opened 2014-07-09
Emscripten no name functions


valls:
Hi!

After running emscripten.sh generates libsodium-js folder with "includes" (.h files folder) and "lib". inside lib creates libsodium.js. How can I run on browser client? It seems that name functions aren't inside libsodium.js. How can I work with that file?

Thanks!

jedisct1:
Closing this, because Javascript work is now happening here https://github.com/jedisct1/libsodium.js

PR #233, opened 2015-01-12
Modular JS wrapper

BatikhSouri:
Hey @jedisct1,

I'm proposing a first version functional version of the JavaScript wrapper we talked about a while ago. To sum up, here is what you can find in this pull request:

* I've completed the JS wrapping for a rather [wide range](https://github.com/LockateMe/libsodium/tree/master/dist-build/js-build/symbols) of methods of the library.
* I've updated [emscripen.sh](https://github.com/LockateMe/libsodium/blob/master/dist-build/emscripten.sh) to let it build the JS wrapper after emscripten and its testing are finished.
* I've built a [web page](https://github.com/LockateMe/libsodium/blob/master/test/js/index.html) that lets you test the following set of algorithms:
  * Curve25519
  * Ed25519
  * Ed25519->Curve25519 transition
  * CryptoBox
  * CryptoSecretBox
  * Scrypt
  * Hash (SHA256 & SHA512)
  * HMAC (SHA256, SHA512, and the default crypto_auth with SHA512/256)
* For all the algorithms listed above, the test procedure also runs on [the test vectors](https://github.com/LockateMe/libsodium/blob/master/test/js/test.js) (harvested from the C version)

I hope it suits your expectations and that it will not be too much of a hassle to merge. (A priori, aside the changes in `emscripten.sh`, I don't see where it can go wrong).

I tried to make it easier to use and keep it multi-platform (as it's a fundamental goal of libsodium). Tell me if there is a problem.

PS: For proper attribution, I'd like to add that I probably wouldn't have been able to build this stuff without looking at @salemware input on #178 and @CryptoNoob's version in #211 (FYI it opened my eyes to the fact that C's `long long` parameters are "trans-typed" as 2 Numbers; since JS numbers cannot hold 64 bit integers).


sarciszewski:
> This pull request contains merge conflicts that must be resolved. 

Do you need any help with this? My usual strategy would be something like...

```sh
git remote add upstream https://github.com/jedisct/libsodium.git
git fetch upstream master
git checkout master
git rebase upstream/master
git push --force origin master
```

jedisct1:
Awesome work, Batikh! :+1: 

A rebased version is available in the `libsodium.js` branch in my repository.

We need to make sure that the library compiles with Emscripten. Such a target is also interesting to spot bugs that native code doesn't easily trigger.

But I am not sure that higher-level wrappers (which are required, or else using the library in Javascript is painful) should be part of the same repository.
I'd rather suggest a distinct repository, with the libsodium repository loaded as a submodule. The Javascript repository can even contain precompiled javascript files that people can copy and use directly.

Some remarks about the code:

- `injectBytes()` is allocating memory on the heap. Output buffers are free()d when converted, but other buffers are not, leading to memory leaks until `free_all()` is manually called. Users can call `free_all()` after each call to libsodium, but that a bit annoying, and it gets worse when an exception is thrown. I'm not sure what the best solution to address this is, but maybe the wrappers should always free everything they allocate using `malloc`.
- `uint` cannot be `0`. Maybe it should. For example the key length in `crypto_generichash` can be zero.
- `leftPadding` in `injectBytes()` is not required. libsodium doesn't require left padding, and that was really only designed to be used with pointers in C.
- `to_hex`/`from_hex`: maybe use the libsodium functions (`sodium_bin2hex` and `sodium_hex2bin`) instead. `bin2hex` is constant time and `hex2bin` is more flexible.
- Maybe the `encode_utf8` and `encode_latin1` function names could describe a bit more what the functions do. How about a single `string_to_uint8array` function instead?

BatikhSouri:
Alright. Should I create the repo in question?

Thanks for your remarks, I'm gonna try to make the appropriate changes. (Sorry for the potentially late reply)

jedisct1:
Hi!

I made a quick try here: https://github.com/jedisct1/libsodium.js

Feel free to fork it or create another repository.

Another thing to look at is `/dev/urandom` emulation.
Recent versions of Emscripten provide the device, but if `crypto.getRandomValues()` is not available, it silently falls back to using `Math.random()` which is a terrible idea for generating secret keys.

Maybe it would be better to include the `pre.js.inc` file that throws an Exception in that case.

BatikhSouri:
Alright, noted. I'll try to work on it as soon as I can.

jedisct1:
https://github.com/jedisct1/libsodium.js is almost ready for prime time.

There are still some issues with the generated API documentation regarding return types, and moving to automated testing using something like Mocha would probably be good. But other than that, it looks pretty good and useable. Congrats, Batikh! :+1: 

Closing this ticket as new tickets should probably be opened in the other repository.

BatikhSouri:
Thanks Frank! Congrats to you for cleaning the messy parts of the code, refactoring and improving upon it! I feel like you deserve more credit than I do :+1:

PR #235, opened 2015-01-15
Use bash for NDK make toolchain

mhamrle:
Use bash for NDK make-standalone-toolchain.sh. The ndk recommends bash, android-ndk-r10c does not work without bash.


jedisct1:
Really? This is sad :(

PR #238, opened 2015-02-02
[DO NOT MERGE} split crypto_sign_verify_detached

sa2ajj:
(This is a work in progress and needs to be completed provided the feedback is positive :))

The motivation is very simple: I checked how many public key signatures/verifications I can do on my machine and it turned out that it's almost three times more signatures than verifications.  I checked the SUPERCOP benchmarks and saw that the this is pretty much the case for all platforms listed there.

Then I looked at the code and noticed public key is converted to some internal representation every time a verification is performed.  Using crypto_box example, I split out this conversion in a separate function.  I did not do any benchmarking on my machine yet, as I wanted to hear what would be the opinion about this kind of change.

I believe the missing bits are:
- [ ] prototype in include file
- [ ] updated test suite (??)

(And, yes, I'd appreciate if somebody would tell what beforenm/afternm stand for :))

### DIFF
--- a/src/libsodium/crypto_sign/ed25519/ref10/open.c
+++ b/src/libsodium/crypto_sign/ed25519/ref10/open.c
@@ -10,29 +10,39 @@
 #include "utils.h"
 
 int
-crypto_sign_verify_detached(const unsigned char *sig, const unsigned char *m,
-                            unsigned long long mlen, const unsigned char *pk)
+crypto_sign_verify_beforenm(unsigned char *A, const unsigned char *pk)
 {
-    crypto_hash_sha512_state hs;
-    unsigned char h[64];
-    unsigned char rcheck[32];
-    unsigned int  i;
+    unsigned int i;
     unsigned char d = 0;
-    ge_p3 A;
-    ge_p2 R;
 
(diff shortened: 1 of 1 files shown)


daurnimator:
Typo: `befor` ->`before`

sa2ajj:
Fixed the typo.

ge0rg:
[OT] while studying the Scripture, I got enlightened that "nm" might mean "number of messages":

> The intermediate data computed by crypto_box_before**nm** is suitable for both crypto_box_after**nm** and crypto_box_open_after**nm**, and can be reused for any **number of messages**.

(Emphasis mine)

tarcieri:
`nm` stands for "nonce" and "message"

sa2ajj:
@tarcieri thanks for the clarification.

jedisct1:
Batch signatures can also benefit from precalculating `sha512(sk)`.

My only concern your changes is that `A` is implementation-specific, and unlike everything else, we can't have a macro whose definition is the required size, unless we expose too much internals.

jedisct1:
This might still be implemented, but not that way, especially since another implementation is coming.

PR #239, opened 2015-02-02
Added capability of setting the initial block counter value for XSalsa20.

Cyberunner23:


Cyberunner23:
Never mind... some test are failing.. Investigating.

PR #240, opened 2015-02-02
Added capability of setting the initial block counter value for XSalsa20.

Cyberunner23:
Second try... Made a small typo that broke the tests... and... stuff...


jedisct1:
Thanks!

The `_ic` functions only make sense when block ciphers are used in counter mode, which is what salsa20 and chacha20 are in disguise.

But the `crypto_stream` interface should remain generic enough not to assume that the underlying stream cipher is using a counter.

For this reason, if initializing the counter for xsalsa20 is being added, this should be only exposed within the `crypto_stream_xsalsa20` namespace.

jedisct1:
My bad, I got confused by the definition in `api.h`.

Looks good.

Cyberunner23:
Thanks for the heads up! Want me to clean it up a little?

jedisct1:
It's been merged in a1b3da7dd990a97e0b9e8fe2933b5f2806186c2d

Last thing that requires an update is the documentation.

PR #243, opened 2015-02-07
android ndk requires arch to be arm not armv7

mosin:
In the android-build.sh script, the call the ${ANDROID_NDK_HOME}/build/tools/make-standalone-toolchain.sh passes TARGET_ARCH as armv7 where as around line 117 of make-standalone-toolchain.sh requires ARCH to be {arm,x86,mips,arm64,x86_64,mips64}


jedisct1:
Merged, thanks!

PR #244, opened 2015-02-07
Add PIE build for Android 5.x support

mosin:
Android 5 requires that files support position independent executables otherwise the following error occurs:
error: only position independent executables (PIE) are supported.
Add PIE flags to CFLAGS and LDFLAGS

### DIFF
--- a/dist-build/android5-armv7.sh
+++ b/dist-build/android5-armv7.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+export CFLAGS="-Os -mfloat-abi=softfp -mfpu=vfpv3-d16 -mthumb -marm -march=armv7-a -fvisibility=default -fPIE"
+export LDFLAGS="-rdynamic -fPIE -pie"
+TARGET_ARCH=arm HOST_COMPILER=arm-linux-androideabi "$(dirname "$0")/android-build.sh"


jedisct1:
Quite confused by this new build script. Doesn't Android 5.x run on other architectures as well?

We can revert cdf863c7312d7f3cc359312bb7ea4892f1c06d74 instead.

But PIE apparently needs to be disabled on some Android versions, see https://github.com/optedoblivion/libsodium/commit/99321088972274fb439725e71b1bedcb9ec490fd

What's the rationale for this?

mosin:
You are correct, PIE needs to be disabled on other Android versions which
is why I created a new file. Let me try building it with cdf863c reverted
and see if it executes. I also sent a pull request for dnscrypt-proxy as
the same issue occurs.

mosin:
I tried reverting https://github.com/jedisct1/libsodium/commit/cdf863c7312d7f3cc359312bb7ea4892f1c06d74 but when trying to execute the binary I still get the same error that only PIE are supported. I added the flags from here and it worked http://stackoverflow.com/questions/24818902/running-a-native-library-on-android-l-error-only-position-independent-executab
LOCAL_CFLAGS += -fPIE
LOCAL_LDFLAGS += -fPIE -pie

jedisct1:
I just tried on 5.0.2, without any changes except bumping the platform version to 18 (tried 14 later, with the same result).

Provided that the final executable is linked with `-pie`, it works out of the box. Didn't had to make any changes to sodium or to its compilation flags. It works both when the library is dynamically and statically linked.

jedisct1:
Confirmed with dnscrypt-proxy. Now that the `--disable-pie` flag has been removed, dnscrypt-proxy runs out of the box on Android 5.0.2.

PR #256, opened 2015-04-27
Update NuGet configuration for version 1.0.3

woobs:
`package.nuspec`, `package.targets`, and `package.xml` are generated from this config file, so I didn't update them (thus it may be reasonable to remove them from the repository)


jedisct1:
Thanks!

PR #258, opened 2015-05-08
Use https when linking where possible.

shancat:
Only makes sense to use TLS when linking from a crypto-related library.
Had to change the libsodium documentation link a bit because the
original URL returns the gitbook.org certificate. cr.yp.to doesn't
actually support encrypting HTTP with TLS so we leave it as-is.


jedisct1:
Merged, thanks!

PR #264, opened 2015-05-22
Added patch to obtain random bytes for Chrome NaCl via IRT -- allows …

arbinger:
…build with NaCl SDK toolchain. Here's my `buildnacl.sh` script in full (only tested on Mac):

	#!/bin/bash

	# Done on a Mac
	export CC="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-clang"
	export NM="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-nm"
	export AR="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-ar"
	export AS="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-as"
	export LD="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-ld"
	export RANLIB="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-ranlib"
	export LIBTOOL="$NACL_SDK_ROOT/toolchain/mac_pnacl/x86_64-nacl/usr/build-1/libtool"
	export CPPFLAGS="-D __native_client__"
	export CFLAGS="-D __native_client__"

	./configure --host=nacl
	make -e

	# Not completely sure why I have to do this post Make, but if I don't,
	# functions can't be found when I try to link against libsodium.a.

	$RANLIB `find . | grep 'libsodium.a'`

	find . | grep 'libsodium.a'


jedisct1:
Thanks!

Supporting Native Client would be awesome, however I'm awfully sorry, but this pull request cannot be merged as-is.

- For consistency, please try to match the style of the rest of the source code.
- If you add a new platform, updating the documentation would also be great.
- The build script only works on OSX. Something a bit more universal would be preferred.
- Manually running `ranlib` in order to see the symbols doesn't inspire too much confidence. Maybe the Native Client toolchain has some issues with `-fvisibility=hidden`, but I'd rather try to understand what the root cause is rather than use a workaround found by accident that might break on other platforms or after updating the SDK.
- The Emscripten-specific code in `randombytes.c` was necessary to work around limitations specific to Emscripten regarding statically initialized function pointers. This is probably not required for Native Client. In addition, Native Client's secure RNG provides `NaClSecureRngDefaultGenUint32`, `NaClSecureRngDefaultGenBytes` and `NaClSecureRngDefaultUniform` that perfectly match Sodium's custom RNG API. So, adding a Native Client randombytes implementation, and setting it as a default on this platform would make more sense. And it would allow users to override it with a different implementation if they ever want to. With your proposed pull request, `randombytes_implementation_name()` is wrong, `randombytes_stir()` doesn't work, neither does `randombytes_uniform()` and I'm not sure about `randombytes_close()`. Which leads to the main reason this change cannot be merged yet:
- The test suite has to run and pass, and that has to be part of the build script. Cross-compiling to platforms where the code cannot be run on the build host is an exception, but in the Native Client case, I the test suite can run on the build platform. A wrapper might be required to run Native Client binaries (`sel_ldr_$(uname -m)`?), but that shouldn't be an issue since this is already something we do for the Javascript target.

arbinger:
I think the problem with the `NaClSecureRng*` API is that it's only available in "trusted" code. The untrusted code only has direct access to `nacl_secure_random()` -- which is more or less the same IRT-based code that's currently in this patch (it uses the above RNG API internally).

Given that though, I don't think it would be real hard to implement the `randombytes_uniform()`, `randombytes_random()` et al using `nacl_secure_random()`.

About

_"adding a Native Client randombytes implementation, and setting it as a default on this platform would make more sense"_

do you mean setting up something like `./src/libsodium/randombytes/nativeclient/`, matching the setup of `sysrandom`?

jedisct1:
Sad to hear that `NaClSecureRng*` is only available in trusted code. But as long as the same PRG remains accessible, it's not a big deal.
`randombytes_uniform()` doesn't need to be reimplemented if the system doesn't provide it; you can just pass `NULL` for a default implementation.

But yes, I referred to `./src/libsodium/randombytes/nativeclient/` matching `sysrandom` and `salsa20`, that only gets compiled on Native Client.

arbinger:
Wasn't sure where `nacl-test-wrapper.sh` should go, decided on `test/default/`. If that's wrong let me know.

Also, I'm happy to update documentation for builds, but is that part of some other repository somewhere?

Basically you need to setup the build environment and pass some params to `configure`. Like so:

```
export CC="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-clang"
export NM="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-nm"
export AR="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-ar"
export AS="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-as"
export LD="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-ld"
export RANLIB="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-ranlib"
export LIBTOOL="$NACL_SDK_ROOT/toolchain/mac_pnacl/x86_64-nacl/usr/build-1/libtool" 
export PNACL_TRANSLATE="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-translate"
export PNACL_FINALIZE="$NACL_SDK_ROOT/toolchain/mac_pnacl/bin/pnacl-finalize"

./configure --host=nacl NATIVECLIENT=1

make && make check
```

jedisct1:
Finally merged ;)

What would you recommend in order to make the build script work out of the box on any system, not just OSX?

arbinger:
Great to hear, thanks. :)

Building on any Autotools capable systems (*nix basically), I think the build would more-or-less be the same as my most recent OSX incantation above.

First you'd need to get the [NaCL SDK](https://developer.chrome.com/native-client/sdk/download) installed so you can establish `$NACL_SDK_ROOT`. Then you'll have to override the default build tools to point there --

```
export CC="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-clang"
export NM="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-nm"
export AR="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-ar"
export AS="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-as"
export LD="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-ld"
export RANLIB="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-ranlib"
export PNACL_TRANSLATE="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-translate"
export PNACL_FINALIZE="$NACL_SDK_ROOT/platform/specific/path/bin/pnacl-finalize"
```

(When you run `./naclsdk update` it sets up the `platform/specific/` part of the paths above, I believe. You just need to determine where the `pnacl-*` tools live after that, and apply.)

Then do `configure` with the below params and then `make`:

```
./configure --host=nacl NATIVECLIENT=1
make && make check
```

Not entirely sure how the build would work on Windows. I'm happy to send a patch to the official docs, I just wasn't sure how to go about doing that.

PR #265, opened 2015-06-01
Add simple PKI certificate format

erayd:
Simple PKI certificate container, bundling from / until validity
timestamps, certificate pk, signer pk, &signature. Allows for
arbitrary additional data to also be attached to the cert and signed /
verified along with it.

Also includes a version number and an 8-bit 'flags' field to allow
room for future revision without breaking backwards compatibility.

Unsure whether this is something you want in libsodium, but it's certainly something I would find extremely useful. If it's something you're open to including, I'm intending to build a 'secure channel' feature on top of it.

If you'd prefer not to merge this kind of thing, then no worries - I just want to check before I sink too much time into developing stuff on top of it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ test/default/box_easy
 test/default/box_easy2
 test/default/box_seal
 test/default/box_seed
+test/default/cert
 test/default/chacha20
 test/default/core1
 test/default/core2
--- a/src/libsodium/Makefile.am
+++ b/src/libsodium/Makefile.am
@@ -25,6 +25,7 @@ libsodium_la_SOURCES = \
 	crypto_box/curve25519xsalsa20poly1305/ref/before_curve25519xsalsa20poly1305.c \
 	crypto_box/curve25519xsalsa20poly1305/ref/box_curve25519xsalsa20poly1305.c \
 	crypto_box/curve25519xsalsa20poly1305/ref/keypair_curve25519xsalsa20poly1305.c \
+	crypto_cert/crypto_cert.c \
 	crypto_core/hsalsa20/ref2/core_hsalsa20.c \
 	crypto_core/hsalsa20/core_hsalsa20_api.c \
 	crypto_core/hsalsa20/ref2/api.h \
--- a/src/libsodium/crypto_cert/crypto_cert.c
+++ b/src/libsodium/crypto_cert/crypto_cert.c
@@ -1,0 +1,146 @@
+
+#include "crypto_cert.h"
+
+#define crypto_cert_HEADER_BYTES (sizeof(uint8_t) /*version*/ \
+                                 +sizeof(uint8_t) /*flags*/ \
+                                 +sizeof(uint32_t) /*valid from*/ \
+                                 +sizeof(uint32_t) /*valid until*/ \
+)
+
+static int
+crypto_cert_swapl(uint32_t n)
+{
+#if defined(NATIVE_LITTLE_ENDIAN)
+    unsigned char *np = (unsigned char*)&n;
+    return
+        ((uint32_t)np[0] << 24) |
+        ((uint32_t)np[1] << 16) |
+        ((uint32_t)np[2] << 8) |
+        (uint32_t)np[3];
+#endif
+    return n;
+}
+
+int
+crypto_cert(unsigned char *out, uint8_t flags, time_t valid_from,
+            time_t valid_until, const unsigned char *ad,
+            unsigned long long adlen, const unsigned char *pk,
+            const unsigned char *s, const unsigned char *k)
+{
+    //build cert
+    unsigned char *pos = out;
+    const uint8_t version = crypto_cert_VERSION;
+
+    memcpy(pos, &version, sizeof(version));
+    pos += sizeof(version);
+
+    memcpy(pos, &flags, sizeof(flags));
+    pos += sizeof(flags);
+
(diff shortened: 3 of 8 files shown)


jedisct1:
This is neat, and definitely something needed by quite a lot of people.

It's probably better to keep it as a distinct project, though, at least until the features and API stabilize.

Verifying a single certificate is useful, but in actual projects, the complexity comes from having more than one certificate. Specifically, the API should be able to:
- Verify a certificate chain (correctly and safely)
- Deal with multiple certificates that might be valid at the same time, which is required for algorithm agility and for key rotation. This requires adding a least a serial number and an algorithm id to each certificate.

erayd:
Separate project is no problem.

Noting your comment: "at least until the features and API stabilise", does this mean you're open to including it in libsodium at some point? If so, then it might make more sense for me to maintain it as a fork of libsodium until such a time as you're happy with where it's at and willing to merge it. If it's something you would prefer to stay a separate project permanently, that makes a significant difference in what I do (e.g. matching libsodium's coding style as per this pull request, rather than using my preferred, what features to include etc).

Regarding what the API needs to be able to do, I had already intended to do chaining - what I have so far is just a first step, and I wanted to check with you about the possibility of inclusion in libsodium before going too much further with it.

Algorithm agility is not something I had planned (because my understanding was that libsodium's algorithm choices were static by design and would not change in the future), however if that's not the case then adding this capability would certainly make a whole lot of sense.

Can you give me some idea of how you see this kind of thing fitting with libsodium? Once I have a better idea of whether or not you're likely to merge it, then I can decide how best to code it and where to document it.

cfcs:
My failing eyesight made me unable to tell your one-letter variables apart, and the long and prosaic "Ulysses"-resemblant comments left me no wiser as to whether this is already included, but, sarcasm aside, wouldn't it be nice to have a signature from the pk being signed included in the certificate to "prove" that the key wants to belong to the certificate?

It could simply be another `sign_detached` signature made with the `certificate sk` key in addition to the one made using the `signer sk`.

This is of course a pretty generic format, but I think many applications would benefit from a feature like that, requiring the certificate holder's consent as to how long the key should be valid and so on?

erayd:
@cfcs: The data is hashed first, because it lives in two different places (the certificate data, and the additional data), and `crypto_sign_detached()` cannot handle multi-part messages, while `crypto_generichash` can.

Regarding your point on concise variable names and minimal commenting, I was simply following the coding style I observed elsewhere in the library. My preference would be to include much more detail in comments, and to use more descriptive variable names - however libsodium isn't my project, and the code is reasonably self-explanatory, so matching style seemed like the better option.

In terms of requiring the signee's consent for what is signed, I like that idea a lot - it's not something I had considered, and isn't something I need personally, but it makes a ton of sense. I'll add it - thanks for your suggestion :-).

cfcs:
Perhaps an API for signatures on arbitrary-length data would be a good idea; certainly seems useful for scenarios like the present, and perhaps for cases larger files have to be signed. @jedisct1, what are your thoughts on this?
Either way, I think it's worth sacrificing some time on copying + clearing the relatively small buffer needed for this data if it saves 32 bytes and a dependency on `blake2b`.

I personally prefer rigorous comments and longer variable names, but no worries -- it's not my project either, so never mind me! ;)

Awesome, thanks for your code! :+1:

erayd:
@cfcs One other thought on the hashing - if the additional data is hashed separately, it allows room for an API that generates a CSR independently of the additional data. So a certificate could be signed and verified without needing the additional data to be present (by signing the hash), which could be convenient if the additional data is very large. Of course, there's nothing to stop the signer from insisting on seeing that data and verifying the hash before signing the certificate, but it seems to me that keeping the hash dependency allows a bit more flexibility.

Re saving 32 bytes - not sure where you're seeing this saving coming from? In the code as it stands currently, the hash isn't included in the certificate, it's just used as an aggregation mechanism (and is generated again when necessary to verify the signature is valid).

cfcs:
Ah, alright! I don't have experience with C, so I was only halfway kidding when I said that I had trouble reading the code - and apparently I did have, I totally missed that the hash was not included :-) 
With that fact in mind, I think it seems like a reasonable design and concur with your points above.

erayd:
Mmm, I find myself wanting to revise a few things in light of your "signee consent" idea. It seems like a handy thing to have, and easy enough to add, so...

jedisct1:
@erayd It can certainly be included in Sodium at some point. Until then, you don't even need to fork it; it can be independant code that one has to link in addition to Sodium, similar to Blobcrypt.

Adding support for multi-part messages to `crypto_sign()` is something I can quickly add to the library, but there's nothing wrong with pre-hashing using blake2 while sketching the certificate format.

Additional data be critical application-specific data that a generic signer may not be aware of. I'd rather enforce the AD to be always considered an integral part of the certificate, so that the only option is to have everything signed/verified. Flexibility is great, but unless some actual scenario absolutely need it, it can only open the door to vulnerabilities.

erayd:
Thanks - in that case, I'll write it with eventual merging in mind.

Multi-part `crypto_sign_detached()` would be awesome if you're willing to add it!

This is your library, so if you'd prefer that the AD be integral and non-separable by design, that's fine :-).

erayd:
Closing PR, will continue development in erayd/crypto_cert and discuss merging with @jedisct1 at a later date.

PR #284, opened 2015-07-21
Zero the buffer used to get poly1305 key for chacha aead

yarrick:
The key for Poly1305 is created by encrypting an empty buffer
with the chacha20 key and supplied nonce. Set the buffer to
zero to be sure that the result is correct no matter the stack contents.

All the other implementations of chacha20-poly1305
I have seen do this.

### DIFF
--- a/src/libsodium/crypto_aead/chacha20poly1305/sodium/aead_chacha20poly1305.c
+++ b/src/libsodium/crypto_aead/chacha20poly1305/sodium/aead_chacha20poly1305.c
@@ -50,6 +50,7 @@ crypto_aead_chacha20poly1305_encrypt(unsigned char *c,
 #endif
 /* LCOV_EXCL_STOP */
 
+    sodium_memzero(block0, sizeof block0);
     crypto_stream_chacha20(block0, sizeof block0, npub, k);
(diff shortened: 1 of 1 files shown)


yarrick:
This is required to get the correct result for me (both on Linux and bare-metal ARM). Let me know if I missed something, I don't really understand how it could have worked before this change.

jedisct1:
Hi,

`block0` is replaced with the output of the stream cipher and its previous content is irrelevant. We call `crypto_stream` here, not `crypto_stream_xor` (which would indeed need zeroing).

To make sure, I filled `block0` with junk where you added the `memzero()` call and kept getting the same output.

When running the test suite, do the tests for tis operation fail?

If not, can you provide a trimmed-down version of your code that exhibits this behavior?

yarrick:
Sorry, my mistake. I see now that there is a memset of the buffer inside the chacha20 function, and also that my test still fails even with this change. Will keep digging, sorry for the noise.

jedisct1:
No problem. And a good opportunity to add a couple tests for this ( c8e898343086265843dae9ac8416cbab34465e98 ); it can't hurt.

PR #285, opened 2015-07-26
Fix VS common props for MT release dll, remove dead code.

evoskuil:


jedisct1:
Merged, thanks!

evoskuil:
NP. Nice job maintaining the VS projects and adding in VS2015. I ran my code generation and came up with no differences.

PR #287, opened 2015-08-02
Fix _FORTIFY_SOURCE redefined issue

wongsyrone:
If we already defined `CFLAGS=-D_FORTIFY_SOURCE=1`, it will still check `-D_FORTIFY_SOURCE=2` and add it to CPPFLAGS. In this case, we redefine `_FORTIFY_SOURCE` to both 1 and 2 if compiler accepts `-D_FORTIFY_SOURCE=2`.

Fix this issue by checking whether `_FORTIFY_SOURCE` was defined or not. If defined, do nothing. If not, then check `-D_FORTIFY_SOURCE=2`.

BTW, OpenWrt's buildroot uses `-D_FORTIFY_SOURCE=1` for most targets by default.


jedisct1:
Merged, thanks!

wongsyrone:
Thanks. :+1:

PR #293, opened 2015-09-04
improve test for unaligned access

mvduin:
ARMv6 and later usually allows unaligned access, although this actually depends on target config. Rather than hardcoding the value on ARM, I've added a test that checks whether the compiler itself thinks unaligned access is supported (which depends on toolchain defaults and can be overridden with -m(no-)unaligned-access )

I've also rephrased the message printed to use AC_MSG_CHECKING


jedisct1:
Looks good!

Thanks for your contribution.

PR #296, opened 2015-09-13
Make libsodium build on CloudABI

EdSchouten:
CloudABI is a purely capability-based runtime environment. More details:

https://github.com/NuxiNL/cloudlibc

Almost all of libsodium builds out of the box, except the RNGs. The following two patches make all of libsodium build on CloudABI as well.

### DIFF
--- a/src/libsodium/randombytes/salsa20/randombytes_salsa20_random.c
+++ b/src/libsodium/randombytes/salsa20/randombytes_salsa20_random.c
@@ -61,7 +61,7 @@ typedef struct Salsa20Random_ {
 static Salsa20Random stream = {
 #if !defined(_MSC_VER) && !defined(__CloudABI__)
     SODIUM_C99(.random_data_source_fd =) -1,
-    SODIUM_C99(.getrandom_available =) 0
+    SODIUM_C99(.getrandom_available =) 0,
 #endif
     SODIUM_C99(.rnd32_outleft =) (size_t) 0U,
     SODIUM_C99(.initialized =) 0,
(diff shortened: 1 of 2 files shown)


jedisct1:
Does `arc4random_*` keep its security properties in a `chroot()`ed environment on all the platforms supported by CloudABI?

jedisct1:
By the way, your diff breaks compilation everywhere (except on CloudABI, maybe) : https://travis-ci.org/jedisct1/libsodium/builds/80088694

EdSchouten:
So the interesting thing is that CloudABI simply does not understand the concept of `chroot()`. There is no global filesystem namespace. It is only possible to use functions like `openat()` to access files underneath a given directory. CloudABI is similar to FreeBSD's Capsicum:

https://www.cl.cam.ac.uk/research/security/capsicum/
https://www.cl.cam.ac.uk/research/security/capsicum/papers/2010usenix-security-capsicum-website.pdf

CloudABI's `arc4random()` is guaranteed to work, regardless of the environment in which the process is started. It also deals with forks properly.

jedisct1:
Looks like at least on FreeBSD, `arc4random()` is [still using RC4](https://svnweb.freebsd.org/base/head/lib/libc/gen/arc4random.c?revision=268642&view=markup#l192). That is a bit embarrassing.

Is that actually the only way to generate random numbers on CloudABI?

EdSchouten:
On CloudABI we currently use Spritz, but work is under way to switch it over to Salsa20.

Calling into the `arc4random()` API only way random data can be obtained, as the underlying system call (`cloudabi_sys_random_get()`) is not accessible outside of the C library itself.

jedisct1:
Merged, with some changes.

`randombytes_salsa20` is not thread-safe, so I don't think we should introduce a dependency on pthreads that would suggest it is.

Since CloudABI and OpenBSD share a usable `arc4random()` implementation, `HAVE_SAFE_ARC4RANDOM` has been added instead of explicitly checking for these environments.

Thanks!

EdSchouten:
The code now builds on CloudABI out of the box. Thanks!

PR #300, opened 2015-09-21
Add makedev check to file descriptor on Linux

paragonie-scott:
I'm not sure what values should substitute the `(1, 9)` pair on other platforms.

### DIFF
--- a/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c
+++ b/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c
@@ -123,6 +123,12 @@ randombytes_sysrandom_random_dev_open(void)
 # if defined(F_SETFD) && defined(FD_CLOEXEC)
                 (void) fcntl(fd, F_SETFD, fcntl(fd, F_GETFD) | FD_CLOEXEC);
 # endif
+# if __linux__
+            if (st.st_rdev != makedev(1, 9)) {
+                errno = EIO;
+                return -1;
+            }
+# endif
                 return fd;
             }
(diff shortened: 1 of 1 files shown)


jedisct1:
I'm not too found of this, for a couple reasons:

* It's system-specific. It's only relevant to Linux kernels, more specifically to old Linux kernels, since on any modern kernel, `getrandom()` is used instead.
* These are hard-coded, magic numbers. They are not defined by the C library and they don't come from a system call either.
* Is it guaranteed that all Linux-derived systems (including Android and all embedded systems) use and will always use these magic numbers? I am not 100% positive. They are free to opt for something different, or completely dynamic. This wouldn't break anything (apps always access devices by their name, since there are no syscalls to access them by major/minor) except libsodium.
* Finally, we have to consider the threat model. An attacker who has the ability to replace a system device with another already completely owns the system.

There are pros and cons. Actually no strong pros and no strong cons. Or maybe a little bit more cons.

paragonie-scott:
> Is it guaranteed that all Linux-derived systems (including Android and all embedded systems) use and will always use these magic numbers? I am not 100% positive. They are free to opt for something different, or completely dynamic. This wouldn't break anything (apps always access devices by their name, since there are no syscalls to access them by major/minor) except libsodium.

Yeah, and that's probably the best point here. This was derived from [A good idea with bad usage: `/dev/urandom`](http://insanecoding.blogspot.co.uk/2014/05/a-good-idea-with-bad-usage-devurandom.html). They also have a table with other values for other OSes.

I wouldn't want to hurt portability.

Issue #308, opened 2015-10-22
Libsodium does not compile for Solaris CC compiler


dtorelliitrsgroup:
I have downloaded the version 1.0.4 and realize that does not compile for Solaris CC compiler due to the following problems:

*export.h*
**__global attribute should be defined as __attribute__(visibility(__global))**
patch for export:
```c
#  if defined(__SUNPRO_C)
#   ifndef __GNU_C__
#    define SODIUM_EXPORT __attribute__(visibility(__global))
#   else
#    define SODIUM_EXPORT __attribute__ __global
#   endif
#  elif defined(_MSG_VER)
#   define SODIUM_EXPORT extern __declspec(dllexport)
#  else
#   define SODIUM_EXPORT __attribute__ ((visibility ("default")))
#  endif
# endif
```

*blake2.h and crypto_generichash_blake2b.h*
**pragma pack only takes one argument for Solaris CC compiler**
patch for pragma push
```c
#ifndef __sun
#pragma pack(push, 1)
#else
# ifndef __GNU_C__
#pragma pack(1)
# else
#pragma pack(push, 1)
# endif
#endif
```
patch for pragma pop
```c
#ifndef __sun
#pragma pack(pop)
#else
# ifndef __GNU_C__
#  pragma pack()
# else
#  pragma pack(pop)
# endif
#endif
```

dtorelliitrsgroup:
I have successfully compile the library with the changes mentioned above...

paragonie-scott:
So, this is the usual process I follow for making a change on Github:

1. Fork the repository. (You can press the "Fork" button at the top of the page.)
2. You should now have a **dtorelliitrsgroup/libsodium** repository.
3. Make your changes there.
4. Visit `https://github.com/jedisct1/libsodium/compare/master...dtorelliitrsgroup:master`
5. Make a pull request by selecting the appropriate branches.

Your workflow can vary depending on if you use a native Git client or the Github website, but the forking step is usually the same.

jedisct1:
This is a bit confusing:

```c
#  if defined(__SUNPRO_C)
#   ifndef __GNU_C__
```

Does that compiler also define `__GNU_C__` ?

jedisct1:
```c
#ifndef __sun
#pragma pack(push, 1)
#else
# ifndef __GNU_C__
#pragma pack(1)
# else
#pragma pack(push, 1)
# endif
#endif
```

Wouldn't

```c
#ifdef __SUNPRO_C
[...]

PR #309, opened 2015-10-22
Libsodium does not compile for Solaris CC compiler #308

dtorelliitrsgroup:
This is a proposal for the fix for #308


jedisct1:
Merged, thanks!

PR #311, opened 2015-10-30
Fix crypto_aead_aes256gcm_decrypt_afternm() when clen < 16

angt:


jedisct1:
Damn, I thought it already had that check.
Good catch. I'm gonna add the related regression test.

PR #320, opened 2015-11-11
Throw the "secure random generator not found" error only on call

BatikhSouri:


jedisct1:
I'm not sure this is a good idea.

With such a change, according to the context, `sodium_init()` returning a success code can either mean "the library is properly initialized, everything is going work as documented", or "the library has been partially initialized, some common functions may not work and will throw an exception later, including those that were expected to never fail".

In the specification, the crypto API is defined within a Worker object, but implementations are still running behind. There is some progress, though, at least in Firefox, so hopefully this won't be an issue for too long.

Meanwhile, if we want the API to be usable within a worker context, the only way is probably by passing messages between the main thread and worker threads. If a worker needs random bytes, it uses `postMessage()` to send the needed number of bytes to the main thread, gets a response later, and resumes its computation.

BatikhSouri:
Alright, I see and understand. And I also thought about passing random bytes from the main thread to the Worker. But as of now, when using libsodium.js in a WebWorker and this message is thrown, doesn't it make libsodium unusable since the `sodium_init()` was interrupted?

jedisct1:
It does. If `sodium_init()` fails, it means that the library is not going to work correctly, which is definitely the case here.
Until web browsers support the crypto API in web workers, `sodium_init()` should only be called in the main thread.
Consuming random bytes from a worker is not going to work with the current code. This requires setting the `getRandomValues()` function to something that sets the event listener once, and then post to the main thread and resumes after having received the bytes. And in the libsodium.js wrapper, this requires adding some helper to set up the event listener of the main thread.

BatikhSouri:
Ok that's what I thought (and I confirmed it in the meantime on iOS 8 and Crosswalk 15; a WebWorker with libsodium failed on the former but worked on the latter). I'm gonna close this pull request and make a custom build of libsodium.js for my special use case.

PR #326, opened 2015-12-03
Add overflow status for nonce utilities

dnaq:
Changed `sodium_add()` and `sodium_increment()` so that they return
-1 on overflow and 0 when overflow hasn't occured.

Even though nonce overflow should be a rare occasion with most primitives in libsodium it is of help to people using the nonce utility functions to be able to find out when it has occured so that correct error handling can be implemented. The reason that it's worth it to modify these functions instead of handling overflow checks outside of the library is to not have to do another pass over the results to find out if overflow has occured.


jedisct1:
I'm against this change.

First, the code might not run in constant time any more.

Then, it only makes sense when using a counter starting at zero, discouraging the use random nonces or with a guaranteed margin between the initial nonce used by each party. What really needs to be detected is wraparounds, relative to the initial vector, not zero. And if this is actually an issue, the protocol itself has a serious design problem.

Finally, the overflow check can easily be achieved using `sodiium_compare()` or `sodium_is_zero()`.

dnaq:
> I'm against this change.
That's perfectly fine, it's your library after all... :)

> First, the code might not run in constant time any more.
Aren't nonces public values? Leaking the content of a nonce should never lead to a security issue. Even if that should be an issue it's solvable by just returning `c` instead (it's already computed by your functions).

> Then, it only makes sense when using a counter starting at zero, discouraging the use random nonces...
For random nonces I believe the usual way is to draw a new random nonce for each message. Agreed on the wraparound relative to an initial nonce, but that would entail saving the initial nonce and doing comparisons against it.
If we take a look at the CurveCP protocol it has 8 byte nonces (with 16-byte prefixes). The protocol is defined such that a connection is closed when the nonces wrap around (which to be honest isn't expected to happen except for extremely long lived connections since 8-bytes is a pretty large value).

> Finally, the overflow check can easily be achieved using sodiium_compare() or sodium_is_zero().

This is true, but would entail another pass over the nonce. Not something that would lead to a large performance hit though.

jedisct1:
The performance hit is negligible and it makes the intent explicit.

dnaq:
Would you be ok with just returning the carry from your functions?

jedisct1:
It's probably better to avoid changing the ABI when not absolutely necessary.

Some versions would return an `int`, other versions would return a `void`, and even for the `libsodium-sys` definitions, this would be a problem.

PR #329, opened 2015-12-10
Fixes #328

### LINKED ISSUE #328: crypto_sign_edwards25519sha512batch_open doesn't return original message
This appears to be a trivial offset error in the function.

Fraser999:
This adds a basic test which exposes the issue and also has a fix for the bug.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,6 +86,7 @@ test/default/core4
 test/default/core5
 test/default/core6
 test/default/ed25519_convert
+test/default/edwards25519sha512batch
 test/default/generichash
 test/default/generichash2
 test/default/generichash3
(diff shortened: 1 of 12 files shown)


jedisct1:
Thanks! The offset was fixed, but this obsolete function will go away, and tests are only for documented functions.

Fraser999:
OK - thanks.  Will you be updating the prebuilt binaries with this fix?

jedisct1:
prebuilt binaries follow the stable branch, so yes. They already include this.

Fraser999:
It looks like libsodium-1.0.7.tar.gz hasn't been updated.

jedisct1:
No, that one will not change.

Issue #333, opened 2015-12-17
crypto_core/curve25519/ref10/*.c perform undefined shifts


bsilver8192:
Many of the files in [crypto_core/curve25519/ref10](https://github.com/jedisct1/libsodium/tree/master/src/libsodium/crypto_core/curve25519/ref10) shift negative integers, which is undefined behavior according to the C standard.

I caught this using Clang's [UndefinedBehaviorSanitizer](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html). Using a libtool which includes [patch 8775](http://savannah.gnu.org/patch/?8775) (I used [master](http://git.savannah.gnu.org/cgit/libtool.git/commit/?id=a5c6466528c060cc4660ad0319c00740db0e42ba), where it's currently the latest commit), I get the following:
```console
$ export CC=clang-3.6 CFLAGS='-fsanitize=undefined -fno-sanitize-recover' LDFLAGS='-fsanitize=undefined -fno-sanitize-recover'
$ ./autogen.sh && ./configure && make && make check
[...]
make[4]: Entering directory '/tmp/libsodium/test/default'
PASS: aead_aes256gcm
FAIL: aead_chacha20poly1305
PASS: auth
PASS: auth2
PASS: auth3
PASS: auth5
PASS: auth6
PASS: auth7
PASS: box
PASS: box2
PASS: box7
PASS: box8
PASS: box_easy
PASS: box_easy2
PASS: box_seal
PASS: box_seed
PASS: chacha20
PASS: core1
PASS: core2
PASS: core3
PASS: core4
PASS: core5
PASS: core6
FAIL: ed25519_convert
PASS: generichash
PASS: generichash2
PASS: generichash3
PASS: hash
PASS: hash3
PASS: onetimeauth
PASS: onetimeauth2
PASS: onetimeauth7
PASS: pwhash
PASS: pwhash_scrypt_ll
PASS: randombytes
PASS: scalarmult
PASS: scalarmult2
PASS: scalarmult5
PASS: scalarmult6
PASS: scalarmult7
PASS: secretbox
PASS: secretbox2
PASS: secretbox7
PASS: secretbox8
PASS: secretbox_easy
PASS: secretbox_easy2
PASS: shorthash
FAIL: sign
PASS: sodium_core
PASS: sodium_utils
PASS: sodium_version
PASS: stream
PASS: stream2
PASS: stream3
PASS: stream4
PASS: verify1
PASS: sodium_utils2
[...]

PR #338, opened 2015-12-17
Fix another shift of a negative number

bsilver8192:
This is the last part of fixing #333.


jedisct1:
Sorry, I have to revert these changes for now.

PR #348, opened 2016-01-16
Add crypt_generichash_blake2b_statebytes function

betafive:
The function crypto_generichash_statebytes exists to dynamically determine the
size of a crypto_generichash_state struct. This is useful when using libsodium
from a language which can't use sizeof on C types. However, no equivalent
existed for the crypto_generichash_blake2b_state struct for users who want to
explicitly use the blake2b algorithm.

The function crypt_generichash_blake2b_statebytes is added to fill this gap.

### DIFF
--- a/src/libsodium/crypto_generichash/blake2/generichash_blake2_api.c
+++ b/src/libsodium/crypto_generichash/blake2/generichash_blake2_api.c
@@ -39,3 +39,10 @@ size_t
 crypto_generichash_blake2b_personalbytes(void) {
     return crypto_generichash_blake2b_PERSONALBYTES;
 }
+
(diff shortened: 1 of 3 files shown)


betafive:
The output of one of the checks looks weird but it reports that it succeeded: https://ci.appveyor.com/project/jedisct1/libsodium/build/1.0.8.14/job/3v9n1ppdlk9q6d9v (says "compile failed" on line 150 then doesn't discover any tests). I'm not sure it that's anything wrong with my change or if it's something dodgy in the auto build/test configuration.

jedisct1:
Merged, thanks!

PR #370, opened 2016-03-09
Add support for One Time Passwords.

95ulisse:
Hi,
this patch adds support for both HOTP and TOTP generation and validation.

This a proposal of the new API for HOTP:

```c
// Generates a new one time password using HMAC-SHA-256
// and puts it in the output buffer `out`.
// `out` must be at least `digits` + 1 long, since this function automatically
// appends the null terminator.
int crypto_hotp_sha256(
    const char* key,
    size_t keySize,
    uint64_t counter,
    unsigned int digits, // Number of digits of the output: 6, 7 or 8 are supported.
    char* out // Output string where the computed value will be saved
);

// Returns 0 if the given password is valid or -1 otherwise.
int crypto_hotp_sha256_verify(
    const char* key,
    size_t keySize,
    uint64_t counter,
    unsigned int digits, // Number of digits of the password: 6, 7 or 8 are supported.
    unsigned int window, // Maximum error range
    const char* m // Password to verify
);

```

And for TOTP is similar:

```c
// It's similar to the above, but has the following parameters in addition:
// `now`: the time at which the calculation should occur
// `initialTime`: the Unix time to start counting time steps
// `step`: width of a single time slice
// In terms of the parameters defined by RFC 6238, `initialTime` is `T0` and `step` is `X`.
int crypto_totp_sha256(
    const char* key,
    size_t keySize,
    uint64_t now,
    uint64_t intialTime,
    uint64_t step,
    unsigned int digits,
    char* out
);

int crypto_totp_sha256_verify(
    const char* key,
    size_t keySize,
    uint64_t now,
    uint64_t initialTime,
    uint64_t step,
    unsigned int digits,
    unsigned int window,
    const char* m
);

```

To abstract away from the low level functions, there are other functions to wrap around the previous ones. `crypto_{h|t}otp` uses HMAC-SHA-512 as a default, and the TOTP version uses the beginning of the Unix epoch (`0`) as initial time and 30 seconds as the time slice. In the same manner, `crypto_{h|t}otp_verify` functions exist.

All the functions exist in both HMAC-SHA-256 and HMAC-SHA-512 flavors, but not HMAC-SHA-1.


jedisct1:
Hi,

Thank you for writing this HTOP/TOTP implementation!

However, this API is very application-specific, while sodium focuses more on the crypto primitives.

This should better live as a distinct library, which will give it more freedom to implement other common requirements, such as QR code generation, Google Authenticator compatibility, additional authenticated data and recovery codes.

It looks like there are a few issues with the current code that can be easily fixed. The numbers don't have a uniform distribution. If you are concerned about side channels and dangling secrets, you may also want to avoid `sprintf()` (and more generally `stdio` completely).

PR #373, opened 2016-03-18
Introduce C++Builder compatibility

jcolli44:
Add new preprocessor directives to allow libsodium to be easily
built in C++Builder.

### DIFF
--- a/src/libsodium/randombytes/randombytes.c
+++ b/src/libsodium/randombytes/randombytes.c
@@ -17,6 +17,10 @@
 # include "randombytes_nativeclient.h"
 #endif
 
+#if defined(random)
+#undef random
+#endif
+
 #ifndef __EMSCRIPTEN__
 #ifdef __native_client__
 static const randombytes_implementation *implementation =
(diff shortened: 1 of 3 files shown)


jedisct1:
Hi, and thanks for these changes!

How do you compile and run the test suite using C++Builder?

jcolli44:
The way I did it was create a console application for each test under a group project. Then I built all console applications, and copied the generated .exe's to /test/default and wrote a batch script to execute each one and print failed if process returned 1, like wintest.bat does.

jedisct1:
Thanks!

I'm not very familiar with Borland C++. Just downloaded it, and I'm stuck trying to compile a Hello World :)

Can you add the missing pieces such as the Makefile to build the library and the required scripts to run the tests?

jcolli44:
Instead of creating a makefile, we ended up creating a bunch of project files (one for the library, and one per test executable) to build libsodium and its test suite in C++ Builder. Since there is already a builds/msvc, we could create a builds/cbuilder directory and put the project files in there.

jedisct1:
Thanks!

The changes have been merged!

But supporting C++Builder seems like a lot of work and the compiler is not free. So don't consider this compiler as supported. New versions might break.

jcolli44:
Ok, thank you for merging our changes!

PR #376, opened 2016-04-04
Fix up relative includes of sodium/common.h

dwrensha:
These includes look like they are intended to be resolved relative to their containing files. Currently, however, they successfully resolve only because the compiler gets the flag `-I./include/sodium` and appending these paths onto `./include/sodium` just happens to do the right thing.

### DIFF
--- a/src/libsodium/crypto_aead/chacha20poly1305/sodium/aead_chacha20poly1305.c
+++ b/src/libsodium/crypto_aead/chacha20poly1305/sodium/aead_chacha20poly1305.c
@@ -10,7 +10,7 @@
 #include "crypto_verify_16.h"
 #include "utils.h"
 
-#include "../../sodium/common.h"
+#include "../../../sodium/common.h"
 
 static const unsigned char _pad0[16] = { 0 };
 
--- a/src/libsodium/crypto_core/hsalsa20/ref2/core_hsalsa20.c
+++ b/src/libsodium/crypto_core/hsalsa20/ref2/core_hsalsa20.c
@@ -8,7 +8,7 @@ Public domain.
 #include <stdlib.h>
 
 #include "crypto_core_hsalsa20.h"
-#include "../../sodium/common.h"
+#include "../../../sodium/common.h"
 
 #define ROUNDS 20
 #define U32C(v) (v##U)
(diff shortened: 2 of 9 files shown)


kentonv:
Can these just be `#include "sodium/common.h"` instead?

It shouldn't make a difference for libsodium, but it makes the code more ekam-friendly (Sandstorm's crazy build system... long story).

jedisct1:
Thank you!

It could reside in `include/sodium/` without being installed, but the definitions in this file are only supposed to be used internally, so I'm not sure it's a good idea to have that file at the same location as public headers.

jedisct1:
Are relative paths acceptable, or is it really a problem when using Ekam?

dwrensha:
https://github.com/jedisct1/libsodium/commit/3927cad829156038a6baa9821d041b76b2f8f0f8 does make things significantly easier for Ekam. (See https://github.com/sandstorm-io/sandstorm/pull/1758)

jedisct1:
To be honest, I'm not a big fan of that change.

But if it makes things easier for Ekam, let's keep it that way!

dwrensha:
For what it's worth, before I noticed the "include/sodium/private" change, I was planning to submit the following commit, which also would have made libsodium more Ekam-friendly: https://github.com/dwrensha/libsodium/commit/9500e92a2498a8ea2407950c1135c77b9fedafaf. It relies on the assumption that the compiler receives a `-I.` flag.

jedisct1:
That would be even better! 

But I don't know if this wouldn't break projects using Cocoapods.

kentonv:
Thanks @jedisct1!

FWIW, what's going on here is:
- We use Ekam as our build system. It's my own crazy design which automatically figures out how to build C/C++ code by looking at the code itself, without consulting anything like a Makefile. It also automatically decides what objects to link into a binary based on symbol tables. It's nice because when we add new source files, they automatically get built and added to all the necessary output binaries without editing additional files.
- We like to have all our main dependencies built by Ekam too, in one big build, so that we don't have to rely on the local system's installed copy nor try to merge a variety of external build systems. (We don't have many dependencies: Cap'n Proto, libsodium, libseccomp.)
- Ekam likes to see everything as one big source tree organized to match the paths seen in `#include` directives.
- So, we bring in the particular libsodium source files we need via a symlink farm.

This has worked surprisingly well, actually. The libsodium source code has kept its inter-file dependencies very clean such that it's actually possible to bring in only the algorithms we need, which I like a lot.

I understand, of course, that you never intended people to build your code this way, so we take full responsibility for any breakages. But where we can find a compromise like this, it helps.

Thanks!

PR #390, opened 2016-05-04
Stable branch for Android 64

alexkeramidas:
Simply changing the platform to 24 seems to do the job for all Android based architectures.
Tested with Android devices and emulators down to Version 16

### DIFF
--- a/dist-build/android-build.sh
+++ b/dist-build/android-build.sh
@@ -24,7 +24,7 @@ export PATH="${PATH}:${TOOLCHAIN_DIR}/bin"
 
 rm -rf "${TOOLCHAIN_DIR}" "${PREFIX}"
 
-bash $MAKE_TOOLCHAIN --platform="${NDK_PLATFORM:-android-16}" \
+bash $MAKE_TOOLCHAIN --platform="${NDK_PLATFORM:-android-24}" \
     --arch="$ARCH" --install-dir="$TOOLCHAIN_DIR" && \
 ./configure \
     --disable-soname-versions \


jedisct1:
Hi Alex,

Unfortunately my Android knowledge is fairly limited, but aren't apps supposed to stick to platform 16 in order to work on Jelly Bean and beyond?

jedisct1:
Note that you can change the target platform just by setting an `NDK_PLATFORM` environment variable prior to running the build scripts. `android-16` is just the default value.

Changes going to the `stable` branch are backports of minor changes initially made in the development branch, that don't change the API/ABI and won't break existing apps and scripts.
So if the default Android platform version is changed, this cannot be done in that branch.

alexkeramidas:
I am testing this for the last three months. There seems to be no problem with changing the NDK platform version as it just adds additional platforms (basically x64 plaforms) and makes it more up to date with modern devices.

On the contrary changing the SDK platform breaks compatibility as you said but this refers to the Java bindings. (Work for this is in this repo https://github.com/joshjdevl/libsodium-jni)

If you think changing the stable branch will break things then is not possible then a fork of it (during the compilation you strongly suggest to use the stable branch) would be a best solution as 64bit devices are coming out already and native android library support for them is kind of mandatory.

alexkeramidas:
Setting the NDK env var according to projects needs is a solution. To work that way you need to do the process 3-4 times if I remember correctly.
Another solution is to make different scripts for every other platform that supports a new arch. I started doing it that way, but since live testing proved there was no need for it I proposed this quick fix. 
No problem arises in any Android version (down to 16) by setting it to v24. Everything works smoothly. 

The reason I chose the stable branch is that you suggest using it.

Congrats on your work by the way! Impressive.

jedisct1:
The platform was downgraded last year from `21` to `16` due to incompatibilities with older devices still running JB. I think that was as issue for ShadowVPN.

Is it guaranteed (not "seems to work for me") that an app compiled with API level 24 will work as expected on Jelly Bean? If not, that seems quite dangerous.

But you don't need to change the scripts to use a non-default API level. Just define an `NDK_PLATFORM` environment variable:

```bash
env NDK_PLATFORM=android-24 dist-build/android-armv8-a.sh
```

alexkeramidas:
Misspelled that sorry. "Seems" is pretty dangerous indeed.
I am testing this for the last three months, devices and emulators.
No problems have risen regardless the platform (24 down to 16).
But that's no 100% guarantee that everything works as they should, IMHO.
To dissolve any concerns I think I will have to wait for feedback from
other users also.
Android is a bit strange when jumping from one platform and/or device to
another (sadly I cannot own every possible device).

On Wed, May 4, 2016 at 12:26 PM, Frank Denis <notifications@github.com>
wrote:

> The platform was downgraded last year from 21 to 16 due to
> incompatibilities with older devices still running JB. I think that was as
> issue for ShadowVPN.
>
> Is it guaranteed (not "seems to work for me") that an app compiled with
> API level 24 will work as expected on Jelly Bean? If not, that seems quite
> dangerous.
>
> But you don't need to change the scripts to use a non-default API level.
> Just define an NDK_PLATFORM environment variable:
>
> env NDK_PLATFORM=android-24 dist-build/android-armv8-a.sh
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/jedisct1/libsodium/pull/390#issuecomment-216809496>
>



-- 
BEST REGARDS
------------
Alexandros G. Keramidas
Agias 22, 41221
Larisa
Hellas
Tel/Fax: 0030.2413.013252/0030.2410237779
Mobile: 0030.6979.005.888

jedisct1:
@sonOfRa @JackWink @Prototik @joshjdevl Ping?

joshjdevl:
Hello, thank you for the ping and attention to this discussion.

It is my [understanding](https://github.com/joshjdevl/libsodium-jni/issues/23#issuecomment-217785055) that using a newer toolchain to compile is backwards compatible.

However, what may not be backwards is using newer features or apis, which the toolchain potentially could allow someone to do. Though it is my understanding that in libsodium itself no new api features are being utilized, rather just a newer toolchain is being used to compile the library.

jedisct1:
So, prior to compiling the library, we should run the `configure` script twice, once with API level 16, then with API level 24, and compare the defined macros. This is going to be a little bit tricky and ugly, but that can work.

joshjdevl:
Hello,

Could you please know the next steps to see this merged in? Is the macro comparison something you were planning to perform?

Thanks.

jedisct1:
If you can add the macro comparison, that would be awesome. If not, I'll try to do it when I get some spare time.

jedisct1:
f8cc885b6f7cb8512aabcd49ed3a510d5be76260 should address this.

Thanks!

jedisct1:
Unfortunately, this break things:

https://github.com/jedisct1/dnscrypt-proxy/issues/514

https://stackoverflow.com/questions/28740315/android-ndk-getting-java-lang-unsatisfiedlinkerror-dlopen-failed-cannot-loca

alexkeramidas:
Hm, havent checked this in a while. So I need to come up with a better solution to support 64bits and legacy OS versions.

PR #393, opened 2016-05-12
Rationalize across msvc project versions.

evoskuil:
I just ran my code generator against the visual studio solutions. I'm really amazed at how clean they are. These are some trivial tweaks to make them uniform across 2010/12/13/15.


jedisct1:
Thank you, Eric!

Merged.

PR #396, opened 2016-05-16
Fixing a small documentation typo

langboost:
Very small change...but hey, since I'm looking through the code. :)


jedisct1:
Good catch, thanks a lot!

PR #398, opened 2016-05-17
don't crash on Win32

sneves:
`InitializeCriticalSection` expects a writable `CRITICAL_SECTION` pointer going in, but it was being fed `NULL`.

Additionally, `_sodium_crit_enter` was not actually entering the critical section by calling `EnterCriticalSection`.

Finally, I've reworked the initialization of the critical section so that `InterlockedCompareExchange` only gets called once if initialization is not happening on another thread at the time. This is the least relevant change.


jedisct1:
Thanks a lot, Samuel!

PR #399, opened 2016-05-18
Support madvise() on FreeBSD

blacklion:
FreeBSD have madvise() behaviors equivalent to MADV_DONTDUMP and MADV_DODUMP but with its own names.
Add definitions for these behaviors used in sodium_mlock() and sodium_munlock() if FreeBSD names are found and Linux ones don't.


jedisct1:
Merged, thanks!

PR #400, opened 2016-05-26
Move copied-and-pasted code into a single header file

drbo:
The code that blocks on `/dev/random` on Linux to ensure that
`/dev/urandom` is properly seeded was duplicated in
src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c and in
src/libsodium/randombytes/salsa20/randombytes_salsa20_random.c.  Move it
into a single header file.

Also, update .gitignore to ignore Emacs temporary files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@
 *.status
 *.tar.*
 *~
+.#*
+#*#
 .DS_Store
 .deps
 .dirstamp
--- a/src/libsodium/randombytes/linux-block-on-random.h
+++ b/src/libsodium/randombytes/linux-block-on-random.h
@@ -1,0 +1,27 @@
+# if defined(__linux__) && !defined(USE_BLOCKING_RANDOM)
+# include <poll.h>
+static int
+randombytes_block_on_dev_random(void)
+{
+    struct pollfd pfd;
+    int           fd;
+    int           pret;
+
+    fd = open("/dev/random", O_RDONLY);
+    if (fd == -1) {
+        return 0;
+    }
+    pfd.fd = fd;
+    pfd.events = POLLIN;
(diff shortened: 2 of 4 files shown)


drbo:
Normally I would not consider myself qualified to do anything serious crypto-related, but I highly doubt this introduced any bugs.

jedisct1:
* Does this change implement new features?
No, it doesn't.

* Does it make the code run faster?
No, it doesn't.

* Does it improve clarity?
It doesn't, it actually reduces it. Instead of having a single-file, self-contained implementation, we now have 95% of the code in one file, and 5% were moved to another file.
The Salsa20 implementation is an example to implement a custom RNG using the `randombytes_implementation` interface. It's probably better to keep the example in a single file instead of having to look at some additional, private header file.

* Does it improve reusability?
For the best or for the worst, some people only extract the functions implementing the operations they need instead of including the whole library into their project. For this reason, sodium favors keeping each group of functions self-contained over factoring everything. The obvious exceptions being functions exported in the public API, relatively large code, and very common functions such as accessing unaligned data.
This change doesn't improve this, but actually makes it slightly worse.

* Does it matches the style of the existing code?
Having whole functions in header files is okay in C++, but very uncommon in C. This is certainly not something done elsewhere in the library, excepted for inlined functions.

* Does it work?
This pull request hasn't been tested. It merged, the package would not compile any more. [Travis complains](https://travis-ci.org/jedisct1/libsodium/builds/133017009) about it on all platforms: `'../linux-block-on-random.h' file not found`, maybe because the file wasn't added to the Makefiles.

This short function was intentionally duplicated in order to keep things simple, clear and easy to reuse.
So let's keep it that way.

PR #405, opened 2016-06-15
document why RtlGenRandom is used

azet:
I took me a while to figure out why `libsodium` uses `RtlGenRandom` over the `CryptGenRandom` functionality on Windows systems. I think this should be properly documented.


jedisct1:
Merged, thanks!

azet:
That was fast :)

PR #417, opened 2016-07-24
Genereate pdb files for VS2010, 2012, 2013

Jan-E:
@jedisct1 The VC14 (VS2015) project file already generates a libsodium.pdb for the libsodium.dll. This PR ensures that the same is done for VS2010, 2012, 2013.


jedisct1:
Merged, thanks!

PR #418, opened 2016-07-24
Fix x64 builds for MSVC < vc11

Jan-E:
@jedisct1 x64 builds for VC9 and VC10 are broken now. I used this patch to create x64 builds for PHP 5.3 and PHP 5.4 (which some people are actually using):
https://www.apachelounge.com/viewtopic.php?t=6359

### DIFF
--- a/src/libsodium/crypto_generichash/blake2/ref/blake2b-compress-avx2.c
+++ b/src/libsodium/crypto_generichash/blake2/ref/blake2b-compress-avx2.c
@@ -7,7 +7,7 @@
 #include <string.h>
 
 #if (defined(HAVE_AVX2INTRIN_H) && defined(HAVE_EMMINTRIN_H) && defined(HAVE_TMMINTRIN_H) && defined(HAVE_SMMINTRIN_H)) || \
-    (defined(_MSC_VER) && (defined(_M_X64) || defined(_M_AMD64)))
+    (defined(_MSC_VER) && (defined(_M_X64) || defined(_M_AMD64)) && _MSC_VER > 1600)
 
 #pragma GCC target("sse2")
 #pragma GCC target("ssse3")
--- a/src/libsodium/crypto_generichash/blake2/ref/blake2b-ref.c
+++ b/src/libsodium/crypto_generichash/blake2/ref/blake2b-ref.c
@@ -429,7 +429,7 @@ blake2b_pick_best_implementation(void)
 {
 /* LCOV_EXCL_START */
 #if (defined(HAVE_AVX2INTRIN_H) && defined(HAVE_TMMINTRIN_H) && defined(HAVE_SMMINTRIN_H)) || \
-    (defined(_MSC_VER) && (defined(_M_X64) || defined(_M_AMD64)))
+    (defined(_MSC_VER) && (defined(_M_X64) || defined(_M_AMD64)) && _MSC_VER > 1600)
   if (sodium_runtime_has_avx2()) {
     blake2b_compress = blake2b_compress_avx2;
     return 0;


Jan-E:
Building in my VS2010 Express still does not succeed (missing kernel32.lib because of a non-defined WindowsSDKDir), but you can open the VS2010 libsodium.sln with a higher VC-version and skip the upgrade to test the x64 build of a DynRelease DLL.

jedisct1:
AFK. Reverting for now.

Jan-E:
OK. Test it in due time. It is not urgent (and not really important).

Jan-E:
BTW. The missing kernel32.lib is a quite common problem with VS2010 and x64:
http://stackoverflow.com/questions/15053337/fatal-error-lnk1104-cannot-open-file-kernel32-lib

Whatever workaround you choose for that, V90 and V100 still stumble over AVX2.

jedisct1:
Thanks!

These changes have been merged to the `stable` branch, and updated binaries will follow.

I didn't know that VS2010 Express could compile x64 code at all. Good to know :)
I'm gonna try the methods described in the thread you mentioned. Thanks!

Jan-E:
>These changes have been merged to the stable branch, and updated binaries will follow.

Maybe it is an idea to keep the versions of libsodium (the lib) and libsodium-php in sync. If you release libsodium v1.0.11 now (from git head with my PRs) and libsodium-php with the same v1.0.11 the versions in phpinfo() would be in sync. And for the Pecl maintainers a new version would be a sign to make compile Windows binaries. Contact @weltling to make sure the new libsodium.dll's are used when building the new php_libsodium.dll's

Jan-E:
> I didn't know that VS2010 Express could compile x64 code at all. Good to know :)

The main problem is to provide the good library-directory. In the VS2010-vcxproj there is a reference to '$(WindowsSdkDir)lib\x64' now (DynRelease|x64 > Properties > Configuration Properties > VC++ Directories > Library directories: at the end). In my VS2010 Express installation $(WindowsSdkDir) is not defined, so kernel32.lib connot be found. I could probably remedy this in the Windows registry, but as PHP is not using VC10, I do not use it day-to-day and I never bothered. 

This does the trick as well:
```
- $(WindowsSdkDir)lib\x64
+ C:\Program Files\Microsoft SDKs\Windows\v6.0A\lib\x64
```
Or adding C:\Program Files\Microsoft SDKs\Windows\v6.0A\lib\x64 to the Additional Library Directories of the Linker.

jedisct1:
> Maybe it is an idea to keep the versions of libsodium (the lib) and libsodium-php in sync

That's not very realistic. A new version of the library is not going to be released with no changes just because a bug was fixed in the PHP bindings. And a new version of the PHP bindings is not going to be needed if a new version of the library with no API changes gets tagged.

PECL maintainers should follow the `stable` branch.

Jan-E:
The script that creates new windows binaries runs only at the moment that a new PECL release is announced. Compare http://windows.php.net/downloads/pecl/releases/libsodium/ with https://github.com/jedisct1/libsodium-php/releases

And apparently that script did not know anything about rebuilding libsodium.dll and using the new headers at https://github.com/jedisct1/libsodium/tree/master/src/libsodium/include

Somehow the PECL maintainers have to be told to rebuild libsodium.dll. And stop using D:\php_noncore_deps\build\libsodium-1.0.5\Build\ReleaseDLL\x64 on the Windows build machine.

weltling:
Thanks for the ping, Jan. For vc9 currently no builds are being produced, as all the PHP versions based on it are discontinued. We also never supported vc9 for x64 in PHP. Thus, probably investing much time in the topic isn't worth it.

@jedisct1 if you need a lib upgrade on the PECL Windows build host, please ping. I don't always have time to track everything happening, so as long as ext builds don't fail - they're considered good. But if there's a specific bugfix in the dependency you explicitly want to deliver, a short ping were worth it.

Thanks.

Jan-E:
> Thanks for the ping, Jan. For vc9 currently no builds are being produced, as all the PHP versions based on it are discontinued. We also never supported vc9 for x64 in PHP. Thus, probably investing much time in the topic isn't worth it.

I know ...

> @jedisct1 if you need a lib upgrade on the PECL Windows build host, please ping. I don't always have time to track everything happening, so as long as ext builds don't fail - they're considered good. But if there's a specific bugfix in the dependency you explicitly want to deliver, a short ping were worth it.

@weltling: This thing started with the remark by Padraig Doran that "functions like \Sodium\crypto_pwhash_str() are not available": https://github.com/jedisct1/libsodium-php/issues/46#issuecomment-234328252
So it is not a bugfix, but missing features that make an upgrade of the dependency worthwhile.

jedisct1:
Thanks a lot for chiming in, @weltling !

The PHP extension can be compiled against older versions of the library (I think as old as version `1.0.0`), but newer features won't be available.

Updating the DLLs to the current code from the `stable` branch would be great, if only for the `crypto_pwhash_*` function set.

Precompiled DLLs are also available here https://download.libsodium.org/libsodium/releases/

Jan-E:
@jedisct1 You might be able to build all the DLL's in one go from a VC14 command prompt. Something like (tailored to your system):
```
C:
cd \msys64\home\Frank\src\libsodium\builds\msvc\vs2010
devenv libsodium.sln /Rebuild "DynRelease|Win32"
devenv libsodium.sln /Rebuild "DynRelease|x64"

cd \msys64\home\Frank\src\libsodium\builds\msvc\vs2012
devenv libsodium.sln /Rebuild "DynRelease|Win32"
devenv libsodium.sln /Rebuild "DynRelease|x64"

cd \msys64\home\Frank\src\libsodium\builds\msvc\vs2013
devenv libsodium.sln /Rebuild "DynRelease|Win32"
devenv libsodium.sln /Rebuild "DynRelease|x64"

cd \msys64\home\Frank\src\libsodium\builds\msvc\vs2015
devenv libsodium.sln /Rebuild "DynRelease|Win32"
devenv libsodium.sln /Rebuild "DynRelease|x64"
```
The v100, v110, v120 and v140 in the vcxproj files make sure you will build the DLL's for the right CRT.

jedisct1:
The DLLs are currently built with the `builds/msvc/build/buildall.bat` script.

weltling:
@jedisct1 oh, so i even didn't have to build :) Upgraded libsodium on the build host and redone the last ext release build. Users can just re-download from the same link.

@Jan-E seems i'll be using the official libsodium builds, as long as @jedisct1 kindly provides them :) 

Thanks.

jedisct1:
Awesome!

Thanks a lot to both of you!

PR #419, opened 2016-08-04
Better AVX2 detection for GCC compilers

ivmaykov:
I had a problem building libsodium 1.0.11 on gcc 4.8.1. The issue was with the AVX2 implementation of BLAKE2b - it looked like the configure script decided that AVX2 was supported by the compiler, but then compilation failed with a bunch of errors that looked like this:

  ...
  {standard input}:391: Error: operand type mismatch for `vpxor'
  {standard input}:398: Error: no such instruction: `vpermq $147,%ymm10,%ymm10'
  {standard input}:404: Error: no such instruction: `vpermq $78,%ymm9,%ymm9'
  {standard input}:411: Error: operand type mismatch for `vpaddq'
  {standard input}:418: Error: operand type mismatch for `vpsrlq'
  ...

I simply added one of the unknown instructions to the program used to detect AVX2 support, inside a `#ifdef __GNUC__` block. This fixed the issue.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -398,10 +398,15 @@ AS_IF([test "x$EMSCRIPTEN" = "x" -a "$host_os" != "pnacl"], [
 #endif
 #pragma GCC target("avx2")
 #include <immintrin.h>
-]], [[ __m256i x = _mm256_abs_epi8(_mm256_setzero_si256()); ]])],
+]], [[
+__m256i x = _mm256_abs_epi8(_mm256_setzero_si256());
+#ifdef __GNUC__
+__asm__ __volatile__ ("vpermq $1,%ymm1,%ymm2");
+#endif
+]])],
     [AC_MSG_RESULT(yes)
      AC_DEFINE([HAVE_AVX2INTRIN_H], [1], [AVX2 is available])
-     AX_CHECK_COMPILE_FLAG([-mavx2], [CFLAGS_AVX="-mavx2"])
+     AX_CHECK_COMPILE_FLAG([-mavx2], [CFLAGS_AVX2="-mavx2"])
      AC_MSG_CHECKING(if _mm256_broadcastsi128_si256 is correctly defined)
      AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
 #ifdef __native_client__


jedisct1:
Merged, thanks!

jedisct1:
Hold on.

The AVX2 implementation of the BLAKE2b compression function doesn't use any inline assembly, only intrinsics.

Looks like your compiler is spitting out instructions that the assembler doesn't support. That's a toolchain problem. Maybe we should detect this, but adding a check that doesn't match the actual code at all doesn't sound good.

configure.ac
```diff
@@ -398,10 +398,15 @@ AS_IF([test "x$EMSCRIPTEN" = "x" -a "$host_os" != "pnacl"], [
...
+]], [[
+__m256i x = _mm256_abs_epi8(_mm256_setzero_si256());
+#ifdef __GNUC__
+__asm__ __volatile__ ("vpermq $1,%ymm1,%ymm2");
+#endif
+]])],
     [AC_MSG_RESULT(yes)
      AC_DEFINE([HAVE_AVX2INTRIN_H], [1], [AVX2 is available])
-     AX_CHECK_COMPILE_FLAG([-mavx2], [CFLAGS_AVX="-mavx2"])
+     AX_CHECK_COMPILE_FLAG([-mavx2], [CFLAGS_AVX2="-mavx2"])
```
jedisct1: Good catch!

ivmaykov:
I'm not sure how to diagnose the toolchain problem. What's the expected behavior with using AVX2 intrinsics on a platform that doesn't support AVX2 instructions? Should the intrinsic function fall back to equivalent AVX1 code?

jedisct1:
The AVX2 code shouldn't be compiled at all.

Replacing `AC_COMPILE_IFELSE` with `AC_LINK_IFELSE` may detect the broken toolchain. 

What OS/distro is that?

jedisct1:
Can you try replacing all the `AC_COMPILE_IFELSE` instances with `AC_LINK_IFELSE` (https://gist.github.com/1a1deb71a91b1c0bf6778b665fd9d8de) and see if it then compiles and links successfully?

jedisct1:
Hi Ilya,

The autoconf script detects what the compiler supports, not what the CPU is. AVX2, SSE4.1, SSSE3 and generic implementations are all compiled in, and the actual implementation is then chosen at runtime according to the current CPU. So the same binary works everywhere.

CentOS 6.7 seems to ship with gcc 4.4.7. So, on your system, it's very likely that other gcc versions have been manually compiled, or come from a more recent version of CentOS. Recent gcc versions don't play well with an outdated assembler/linker.

Can you post (as a gist) the `config.log` file you get after running the `configure` script with `AC_LINK_IFELSE`?

ivmaykov:
I think you're right, it's probably an issue with our toolchain. I tracked down the versions of "as" used by the different gcc versions (by looking at build logs and finding what the PATH was set to). The version used by gcc 4.8.1 is old (2011) and fails to compile the following 2-line program:

        vpermq  $1, %ymm1, %ymm2
        ret

The version used by gcc 4.9 and higher succeeds.

So while I agree that this is a bug in our toolchain, it's possible that other people could have the same issue if they have a mismatch between the features supported by their 'gcc' and 'as' binaries. Maybe there's a more elegant way to detect such an issue? Up to you if you want to unmerge the PR or not ...

ivmaykov:
I am digging into it some more. The problem seems to be that some subset of AVX instructions (maybe AVX1?) is supported by the older assembler, and gcc optimizations are eliminating the problematic instructions that should have been generated by the intrinsic function. Maybe it makes sense to compile the test programs that detect AVX features with -O0.

ivmaykov:
Please see https://github.com/jedisct1/libsodium/pull/420

PR #420, opened 2016-08-04
Better AVX2 detection, try 2 (only using intrinsic functions)

ivmaykov:
Here's a second attempt at better AVX2 detection. I dug pretty deeply into the generated assembly, and found that the old version (`__m256i x = _mm256_abs_epi8(_mm256_setzero_si256());`) generated ASM that didn't actually have any AVX2 instructions in it - they were optimized away completely (though I think there were some AVX1 instructions left over). So in cases where the compiler version was recent enough to support AVX2 but the assembler was too old and didn't support it, the configure script would fail to detect the problem.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -398,13 +398,12 @@ AS_IF([test "x$EMSCRIPTEN" = "x" -a "$host_os" != "pnacl"], [
 #endif
 #pragma GCC target("avx2")
 #include <immintrin.h>
-#include <string.h>
 ]], [[
 // the code below is intentionally crafted to make sure the compiler
 // can't optimize the AVX2 instructions away.
-__m256i x = _mm256_set1_epi32(0xdeadbeef);
-__m256i y = _mm256_permutevar8x32_epi32(x,x);
-return memcmp(&x, &y, sizeof(x)) != 0;
+__m256 x = _mm256_set1_ps(3.14);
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -398,7 +398,14 @@ AS_IF([test "x$EMSCRIPTEN" = "x" -a "$host_os" != "pnacl"], [
...
 #pragma GCC target("avx2")
 #include <immintrin.h>
-]], [[ __m256i x = _mm256_abs_epi8(_mm256_setzero_si256()); ]])],
+#include <string.h>
+]], [[
+// the code below is intentionally crafted to make sure the compiler
+// can't optimize the AVX2 instructions away.
+__m256i x = _mm256_set1_epi32(0xdeadbeef);
+__m256i y = _mm256_permutevar8x32_epi32(x,x);
+return memcmp(&x, &y, sizeof(x)) != 0;
```
jedisct1: Does `return _mm256_movemask_ps(_mm256_cmp_ps(x, y, _CMP_NEQ_OQ));` work as well?

ivmaykov:
@jedisct1 yes it does, though I have to cast x and y to __m256. Do you want me to update the diff?

ivmaykov:
Actually I can change the first 2 lines to do operations on __m256 types (float vs. int vectors), then the cast is not needed.

jedisct1:
It contains 42, it has to be the right answer.

Merged, thanks!

PR #424, opened 2016-09-02
configure and make quiet flags

joshjdevl:
Adding --quiet flag to configure and make for Android build to reduce verbose output.


jedisct1:
Why?

joshjdevl:
We build for all architectures. This results in [logs](https://api.travis-ci.org/jobs/157226008/log.txt?deansi=true) that are at least several MBs, in addition to slowing the automated builds.

jedisct1:
It's usually not a bad idea to print what's going on, especially for potentially long operations, or people will think that the process is stuck.

These logs are also immensely useful for debugging.

If you want to avoid them in automated builds, maybe it would be better to just redirect `stdout` to `/dev/null` with `android-build.sh > /dev/null` instead of changing the script itself.

PR #429, opened 2016-09-20
add preprocessor flag to skip blocking /dev/random during libsodium init

robspychala:
Android emulators running on CentOS 6.0 *sometimes* block indefinitely on `/dev/random` access. 

This bug most likely exists in interaction between CentOS 6.0 and Android NDK, but adding this preprocessor and turning off blocking was a simpler fix to work around this problem.

The init code works ok with out the preprocessor defined on OS X and CentOS 7.0 and up.


jedisct1:
Merged, thanks!

PR #436, opened 2016-10-28
add extern C support to make it can be called in C++ code.

renweibo:
These two header files seems also should put into include/sodium directory

### DIFF
--- a/src/libsodium/crypto_pwhash/scryptsalsa208sha256/crypto_scrypt.h
+++ b/src/libsodium/crypto_pwhash/scryptsalsa208sha256/crypto_scrypt.h
@@ -33,6 +33,13 @@
 #include <stdint.h>
 #include <stddef.h>
 
+#ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


jedisct1:
These are internal symbols. Functions defined here are not supposed to be called by external code, even in C.

The internal interface may change, but more importantly, the symbols are not exported. Which is why they are not documented anywhere.

renweibo:
It seems that you want these files internal. It's OK to keep them internal.

Just to make life easy. These header should be consistent with other header files. C header file has "external c" part is friendly to C/C++ project.

FYI, I saw project https://github.com/Novators/libsqrl use these there two header files.

jedisct1:
Exported symbols are only the ones defined `src/libsodium/include`. Other headers are even not installed at all. No precompiled binary packages contain them.

I don't understand why `libsqrl` tries to include `crypto_scrypt.h`. It makes no sense since the public functions are already defined by including `<sodium.h>`

PR #449, opened 2016-11-30
GCC2 bug seeing empty statement at 2 semicolons

R167:
GCC2 breaks when it encounters a double semicolon due to parsing it as an empty statement before the variable declaration. Breaks compiling on Haiku.


jedisct1:
Merged, thanks!

PR #452, opened 2016-12-04
core_salsa20.c: Avoid undefined rotate behaviour

spacekitteh:
http://blog.regehr.org/archives/1063


jedisct1:
`rotate()` is always called with constants, that will always be the same for the salsa20 function.

So `rotate()` will never be called for a `0`-bit rotation. There is no undefined behavior here, and out-of-bounds rotations are systematically checked with ubsan before every release.

Also, calling `assert()` can potentially have a cost (libsodium unconditionally undefined `_NDEBUG`). That wouldn't be good as this function is called very frequently.

PR #456, opened 2016-12-13
fixed Visual Studio 2015 project file

jeongkyu:
This PR fixed VS2015 project file for the recent renaming of box_xsalsa20poly1305 in commit 669ed59.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@
 *.status
 *.tar.*
 *~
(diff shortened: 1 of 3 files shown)


jedisct1:
There have been many other changes to the files, including the addition of all the ones required to support xchacha20.
Visual Studio 2010, 2012 and 2013 solutions would also need to be updated.

PR #459, opened 2016-12-20
Redundancy in curve25519_ref10.c

paragonie-scott:
Unless I'm mistaken, this is kind of redundant.


paragonie-scott:
Okay I have no idea what broke in Windows. I must be mistaken about something.

PR #460, opened 2016-12-23
xchacha20poly1305: be compatible with ietf chacha20poly1305

zx2c4:
Due to SSL, the IETF version of chacha20poly1305 is going to be the one
that's in libraries places. While the 12-byte nonce thing is a little
weird, it has other benefits, like adding padding to the auth tag, which
might help fend off certain attacks.

But more importantly, since chacha20poly1305 in the IETF construction is
lots of places, it would be useful to be able to build xchacha20poly1305
out of it. Fortunately it's very easy to make hchacha20 (either
stand-alone, or out of the normal chacha20 block function), and then
that can be composed with an existing library's chacha20poly1305. It
looks a bit like this:

    xchacha20poly1305(input, key, nonce) {
        new_key = hchacha20(key, nonce)
        return chacha20poly1305(input, new_key, nonce + 16)
    }

This is also an efficient way to do it, since it means hchacha20 must
only be computed once.

Unfortuantely, by pushing your xchacha20poly1305 to use the non-IETF
chacha20poly1305 construction, you deprive virtually all other libraries
that only support the more common IETF construction the ability the
ability to interoperate with libsodium, through the simple construction.
Rather, it forces everyone to reimplement the AEAD part.

So, this commit fixes xchacha20poly1305 to use the IETF construction
with the padding.

### DIFF
--- a/src/libsodium/crypto_aead/xchacha20poly1305/sodium/aead_xchacha20poly1305.c
+++ b/src/libsodium/crypto_aead/xchacha20poly1305/sodium/aead_xchacha20poly1305.c
@@ -12,6 +12,8 @@
 
 #include "private/common.h"
 
+static const unsigned char _pad0[16] = { 0 };
+
 int
 crypto_aead_xchacha20poly1305_encrypt_detached(unsigned char *c,
                                                unsigned char *mac,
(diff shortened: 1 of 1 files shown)


jedisct1:
That's a good idea, but this PR cannot be merged as-is. All the functions have to be renamed, and all the functions have to be updated, not just a random one.

zx2c4:
Yes, right.

Is this something you're planning on taking care of yourself?

zx2c4:
@jedisct1  I think, by the way, that it will be much easier, and more performant, to just define xchacha20poly1305 in terms of chacha20poly1305. It saves a useless extra hchacha20 call on the block0 computation. I'll implement it and resubmit.

PR #461, opened 2016-12-24
xchacha20poly1305: optimize and be compatible with ietf chacha20poly1305

zx2c4:
Due to SSL, the IETF version of chacha20poly1305 is going to be the one
that's in libraries places. While the 12-byte nonce thing is a little
weird, it has other benefits, like adding padding to the auth tag, which
might help fend off certain attacks.

But more importantly, since chacha20poly1305 in the IETF construction is
lots of places, it would be useful to be able to build xchacha20poly1305
out of it. Fortunately it's very easy to make hchacha20 (either
stand-alone, or out of the normal chacha20 block function), and then
that can be composed with an existing library's chacha20poly1305. It
looks a bit like this:

```
    xchacha20poly1305(input, key, nonce) {
        new_key = hchacha20(key, nonce)
        return chacha20poly1305(input, new_key, nonce + 16)
    }
```

This is also an efficient way to do it, since it means hchacha20 must
only be computed once.

Unfortuantely, non-IETF xchacha20poly1305 means that you deprive
virtually all other libraries that only support the more common
IETF construction the ability the ability to interoperate with
libsodium, through the simple construction. Rather, it forces
everyone to reimplement the AEAD part.

So, this commit adds a xchacha20poly1305 that uses the IETF construction
with the padding.

While we're at it, we redefine xchacha20poly1305 in terms of
chacha20poly1305, which gives the same output, but computes one less
hchacha20 and is generally a lot cleaner and simpler to understand.

Signed-off-by: Jason A. Donenfeld <Jason@zx2c4.com>

### DIFF
--- a/src/libsodium/crypto_aead/xchacha20poly1305/sodium/aead_xchacha20poly1305.c
+++ b/src/libsodium/crypto_aead/xchacha20poly1305/sodium/aead_xchacha20poly1305.c
@@ -5,9 +5,8 @@
 #include <string.h>
 
 #include "crypto_aead_xchacha20poly1305.h"
-#include "crypto_onetimeauth_poly1305.h"
-#include "crypto_stream_xchacha20.h"
-#include "crypto_verify_16.h"
+#include "crypto_aead_chacha20poly1305.h"
+#include "crypto_core_hchacha20.h"
 #include "utils.h"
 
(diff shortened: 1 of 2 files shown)


jedisct1:
Merged, thanks!

Issue #457, opened 2016-12-19
Expose ARGON* constants for `crypto_pwhash`


emilbayes:
Is there a reason the `ARGON*` constants are not exposed for `crypto_pwhash`?

I'm specifically thinking of the input restrictions, which you might want to check for when taking user input. Especially if the password hash function might be changed in the future, it'd be nice for completeness to have [min, max] password length, [min, max] hash output length etc. officially exposed instead of hardcoded by the application developer.

jedisct1:
Hi Emil,

Indeed, these would be nice to have. I'm gonna add these as soon as possible.

Thanks!

PR #464, opened 2016-12-26
WIP: crypto_pwhash constants

emilbayes:
I've attemted a PR for #457, and tried to expose a uniform API for both `crypto_pwhash`, `crypto_pwhash_argon2i` and `crypto_pwhash_scryptsalsa208sha256`, however I have some standing issues before this PR is finished:

* I was only able to read off one hard limit for scrypt, namely `OPSLIMIT_MIN` as that's simply overridden if too low
* Due to my inexperience with C, I wasn't able to import the `argon2.h` header file. Should all the limits from there simply be copy/pasted into the `crypto_pwhash_argon2i.h` header file? That seems to have been the case for all other header files, however some constants are resolved at compile time (?) it seems, like the memlimit.

I have a PR for `libsodium-docs` that will follow shortly.

### DIFF
--- a/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c
+++ b/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c
@@ -20,6 +20,30 @@ crypto_pwhash_argon2i_alg_argon2i13(void)
 }
 
 size_t
+crypto_pwhash_argon2i_bytes_min(void)
+{
+    return crypto_pwhash_argon2i_BYTES_MIN;
+}
+
+size_t
+crypto_pwhash_argon2i_bytes_max(void)
+{
+    return crypto_pwhash_argon2i_BYTES_MAX;
+}
+
+size_t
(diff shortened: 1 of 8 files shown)


jedisct1:
Hi Emil,

And thanks a lot for your contribution!

Regarding the Argon2 constants: just hardcode them. Public headers should not depend on private headers, and it's also convenient to just be able to review these public headers and see what the values are.

opslimit/memlimit: the limits really depend on the underlying algorithm. In scrypt, these are not really independent. What they actually define is an overall work factor, based on their product, and the min/max values depend on this product. The max value is also very implementation-specific and may change.
These parameters are independent in Argon2, but since `_pwhash` is a high-level API, we can't assume that this will be always the case.
To make a long story short, maybe we should avoid the {ops,mem}_(min,max} definitions.

emilbayes:
Just wanted to ping here again after RWC :)

jedisct1:
Merged, thanks!

PR #471, opened 2017-01-08
[Findsodium.cmake] Visual Studio generic algorithm

BurningEnlightenment:
I've replaced the manual MSVC version to VS version mapping with two simple functions which should spare us from updating the script every time Microsoft releases a new Visual Studio (except they break their versioning pattern _again_ 😒).


BurningEnlightenment:
@jedisct1 can you merge this, please?

jedisct1:
Done! Sorry, I didn't notice this PR.

Issue #466, opened 2016-12-27
sodium_FOUND not defined


LongShao007:
I use FindSodium.cmake to determine whether the installation of libsodium, but sodium_FOUND not defined, sodium_INCLUDE_DIR and sodium_LIBRARY_RELEASE are defined.
I tried libsodium verison 1.0.11 and master.
cmake version 3.7.1
ubuntu 14.04 64
`set(sodium_FOUND TRUE)`

jedisct1:
Any idea, @BurningEnlightenment ?

BurningEnlightenment:
I'll look into this when I'm back from my ski holidays (you can assign me if you want :wink:)

jedisct1:
Enjoy your skip trip!

BurningEnlightenment:
@LongShao007 `sodium_FOUND` is always defined by [FindPackageHandleStandardArgs](https://cmake.org/cmake/help/v3.7/module/FindPackageHandleStandardArgs.html). But AFAIK it's not a cached variable, i.e. you can't inspect its value via `cmake gui` or `ccmake`. However you should still be able to use it from your script. Have you tried checking whether `sodium_FOUND` has been defined with something like the following?
```cmake
if (NOT DEFINED sodium_FOUND)
    message(FATAL_ERROR "sodium_FOUND wasn't defined")
endif()
```

LongShao007:
i use FindSodium.cmake
`find_package(Sodium)
if(NOT sodium_FOUND)
  message(FATAL_ERROR "Sodium not found!")
endif(NOT sodium_FOUND)`
cmake report error. "Sodium not found!"

BurningEnlightenment:
I'm not sure, whether I understood your problem correctly. 
Do you have problems detecting the sodium libraries, i.e. the three cached variables `sodium_INCLUDE_DIR`, `sodium_LIBRARY_DEBUG` and `sodium_LIBRARY_RELEASE` are not set appropriately? Or are those variables set to _valid_ values, but `sodium_FOUND` is set to "FALSE"?

LongShao007:
my code:

> find_package(Sodium)
message("sodium_FOUND = " "${sodium_FOUND}")
message("sodium_INCLUDE_DIR = " "${sodium_INCLUDE_DIR}")
message("sodium_LIBRARY_RELEASE = " "${sodium_LIBRARY_RELEASE}")
if(NOT DEFINED sodium_FOUND)
[...]

PR #473, opened 2017-01-11
RENAME FindSodium.cmake TO Findsodium.cmake

BurningEnlightenment:
Fixes jedisct1/libsodium#466.


jedisct1:
Merged, thanks! The Makefile also needs an update, but I'll take care of this.

Issue #479, opened 2017-01-23
Wrong file name in crypto_secretbox


MartinDrab:
The file src\libsodium\crypto_secretbox\xsalsa20poly1305\ref\secretbox_xsalsa20poly1305.c seems to be listed within the MSVS2015 solution under wrong name (box_xsalsa20poly1305.c). Either renaming the file, or changing its name in the solution fixed the issue (the library compiles at least). I may try to create a pull request and fix this if necessary. But I belive this is such a minor fix that the authors may fix the problem by themselves.

jedisct1:
Please use releases, or the stable branch.

The development code is not supposed to work at all on Visual Studio yet.

PR #480, opened 2017-01-24
Problem: wrong reference to non-existent file box_xsalsa20poly1305.c …

sigiesec:
…in VS2015 project

Solution: added correct file secretbox_xsalsa20poly1305.c to vcxproj


BurningEnlightenment:
FYI #479

PR #501, opened 2017-03-01
Update comment on crypto_generichash_state allocation with sodium_malloc()

AlexIljin:
I noticed a discrepancy between the PDF and the source code comments in utils.h. This fixes it.


jedisct1:
Merged, thanks!

PR #503, opened 2017-03-05
Packaging libsodium for .NET Core (Update)

ektrah:
* Replaces openSUSE Build Service with Docker
* Updates packaging scripts to .NET Core SDK RC4
* Removes support for fedora.23-x64 and rhel.7-x64
* Adds support for alpine.3-x64


jedisct1:
Awesome!

Thank you so much for the update!

jedisct1:
By the way, 1.0.12 will be released soon.

Does the .NET Core package need to be updated before the release is tagged? After?

What would be the best way to sync up so that the update works flawlessly for .NET users?

ektrah:
The script downloads everything it needs from download.libsodium.org and [Homebrew](https://bintray.com/homebrew/bottles/libsodium); it does not use git. So, once 1.0.12 is released and macOS binaries are available, the .NET Core package can be built and published. The script does not need to be updated for each libsodium release with this pull request.

The dependency on Homebrew is a bit annoying, but I haven't found a good way to build macOS binaries yet. My plan is to watch the libsodium mailing list for release announcements and then check Homebrew for updates every couple of days.

jedisct1:
Thanks!

Travis can build macOS binaries, but the Homebrew solution isn't too bad.
Homebrew maintainers do a stellar job and new releases are available very quickly.

PR #507, opened 2017-03-06
Add missing file for packaging for .NET Core

ektrah:


jedisct1:
Merged, thanks!

ektrah:
For .NET Core, only the dynamic symbol table is needed, which is not stripped. Stripping the other symbols reduces the binary size from ~2 MB to ~500 KB.

jedisct1:
But doesn't that make debugging much more difficult?

ektrah:
Debugging in .NET ends at the managed/native boundary by default. For example, if a C library called from C# dereferences a null pointer, all you get is an AccessViolationException at the line where the C# code calls into the C code, without any information where in the C code the error occurred.

There are some configuration settings in Visual Studio that can be turned on to extend debugging into native code, but I'm not sure how frequently these are used. And this is Visual Studio-only, anyway; the cross-platform .NET Core debugger doesn't seem to have this feature.

PR #519, opened 2017-03-24
Fix up buildbase.bat, VS projects whitespace and remove ARM.props

evoskuil:
The `msvc/properties/ARM.props` file is not the right place for a build customization. The properties directory exists to support the build configurations that are placed into the VS project files. These are purposely very general and the set should really never change.


jedisct1:
Thank you!

Issue #438, opened 2016-11-02
enhancement request:  side-channel safe base64


udf2457:
Hi,

I note you have a side-channel safe bin2hex.

Any chance you could implement similar for base64 ?  ;-)

paragonie-scott:
I started on porting [this from C++](https://github.com/Sc00bz/ConstTimeEncoding) a while ago, but got confused by C's type system.

A PHP implementation of side-channel safe RFC 4648 exists here: https://github.com/paragonie/constant_time_encoding

udf2457:
Well I just get confused by C in general, so I don't blame you !  ;-)

Thanks for the PHP link, looks like just the thing I need.

ivan:
I spotted an implementation of this in https://github.com/WireGuard/WireGuard/commit/8f002a9f2ccbb4365ad1242ce24aa379eb1c0a24

zx2c4:
@ivan I'd appreciate it if you'd update or remove your comment, as it contains a commit that's not actually valid anymore. I went through a flurry of force pushing, and you wound pasting something silly.
--
The real file is here: https://git.zx2c4.com/WireGuard/tree/src/tools/base64.c

PR #524, opened 2017-03-31
Add constant-time Base64 encoding and decoding

entrope:
This should address #438.

I based this on the C++ version by Sc00bz (linked in #438), but took the liberty of handling padding in constant time by reducing b64_len.

This code assumes two's complement representation of signed integers; is that worth avoiding?

### DIFF
--- a/src/libsodium/include/sodium/utils.h
+++ b/src/libsodium/include/sodium/utils.h
@@ -62,6 +62,15 @@ int sodium_hex2bin(unsigned char * const bin, const size_t bin_maxlen,
                    const char ** const hex_end);
 
 SODIUM_EXPORT
+char *sodium_bin2base64(char * const base64, const size_t base64_maxlen,
+                        const unsigned char * const bin, const size_t bin_len);
+
+SODIUM_EXPORT
+int sodium_base642bin(unsigned char * const bin, const size_t bin_maxlen,
+                      const char * const hex, const size_t hex_len,
+                      size_t * const bin_len, const char ** const base64_end);
+
+SODIUM_EXPORT
 int sodium_mlock(void * const addr, const size_t len);
 
 SODIUM_EXPORT
--- a/src/libsodium/sodium/utils.c
+++ b/src/libsodium/sodium/utils.c
@@ -367,6 +367,156 @@ sodium_hex2bin(unsigned char *const bin, const size_t bin_maxlen,
     return ret;
 }
 
+/* Derived from original code by Steve "Sc00bz" Thomas. */
+static void
+base64_encode_chunk(char dest[4], const unsigned char src[3])
+{
+    int part, diff, i;
+    unsigned char chunk[4];
+
+    chunk[0] = src[0] >> 2;
+    chunk[1] = (src[0] << 4) | (src[1] >> 4);
+    chunk[2] = (src[1] << 2) | (src[2] >> 6);
+    chunk[3] = src[2];
+    for (i = 0; i < 4; ++i) {
+        part = chunk[i] & 63;
+        diff = 0x41;
+
+        /* Comments assume ASCII representation, code does not. */
+        /* if (part > 25) diff += 'a' - 'A' - 26; */
+        diff += ((25 - part) >> 8) & 6;
+        /* if (part > 51) diff += '0' - 'A' - 26; */
+        diff -= ((51 - part) >> 8) & 75;
+        /* if (part > 61) diff += '+' - '0' - 10; */
+        diff -= ((61 - part) >> 8) & 15;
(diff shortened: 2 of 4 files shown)


jedisct1:
To be honest, I'm a bit reluctant to include functions for base64 encoding/decoding.

base64 is a mess, with no less than [15 common, but incompatible variants](https://en.wikipedia.org/wiki/Base64#Variants_summary_table).

As an example, a base64 codec is [already present in libsodium](https://github.com/jedisct1/libsodium/blob/c229663acf6083867b8df35aeb8647e6d4db1270/src/libsodium/crypto_pwhash/argon2/argon2-encoding.c#L53).

Can we remove it and replace it with your implementation of a given variant? Unfortunately not, because this is not the same variant (no padding).

Would anyone use your implementation or the existing one to pass around signatures in URLs? No, because another variant would be a better fit.

Base64 is less efficient than Base85. Base64 is not human-readable as Base58.

Essentially, I don't see any valid reason to use Base64. Ever.

entrope:
I am not sure what you mean by "this is not the same variant".  The decoder in this PR should handle padding being present and padding not being present.  The encoder here does not append padding, but could (at least when 

The variants listed on Wikipedia vary by the last two characters used; by whether padding is mandatory, optional, or prohibited; line length; how to handle non-Base64 characters; and checksum.  (The non-Base64 handling and checksum are only really different for one variant in each case.)  I think it would be straightforward to parameterize this while preserving the constant-time aspect by passing a string that specifies the last two Base64 digits, the padding character, and perhaps how to handle whitespace and/or non-Base64 characters.

Base85 also has a plethora of variants, some of which leak content by design (the "z" and "y" abbreviations for btoa and Ascii85) and need escapes for many digits, and others (Z85 and RFC1924/git version) that use a lot of non-contiguous ranges that slow down constant-time conversion.  RFC1924 works on 128-bit chunks; git uses the same digits, but works on 32-bit chunks.

Base58 also has a number of variants, all of which use lots of non-contiguous ranges, and does not have any standard for how many bytes to convert per chunk.  (41 bits in 7 Base58 digits?  64 bits in 11 Base58 digits?  Something else?)

All of the text formats are unhappy mediums: base16 is inefficient, base58 is inefficient and awkward except for human I/O, base64 is widely used for key and certificate exchange but has a lot of variants, base85 has several variants that are more fundamentally different than the base64 variants.

If you want to close this, I have no great objection, but it might be useful to mention your concerns in issue #438 for the next person to come along who is looking to help out.

jedisct1:
This is a sad state of affairs. There are many encoding systems and many variants, each of them being useful.

Ideally, we would want to support them all, but being a BaseXX conversion library is not really the purpose of libsodium.

Still, having great, constant-implementations of these would be very useful, but I think this would be a project on its own.

jedisct1:
Of course we could just implement one system/variant. This is what PHP does in its standard library. 

But if you look at the user comments in that function documentation, you will notice many code snippets performing additional substitutions (which would defeat the point of a constant-time implementation) to construct different variants.

So we can't really offer just one variant, or the same thing is going to happen.

paragonie-scott:
> base64 is a mess, with no less than 15 common, but incompatible variants.

But it has [an RFC](https://tools.ietf.org/html/rfc4648). If this weren't a `wontfix`, I would recommend only implementing the RFC 4648 variants.

> Essentially, I don't see any valid reason to use Base64. Ever.

The only reason to use it is because Base-N (where N is an even multiple of 2) is easy to implement in constant-time.

> Of course we could just implement one system/variant. This is what PHP does in its standard library.

They also refused to make their implementation constant time, fearing FUD about performance concerns, so I [took and expanded Sc00bz's code](https://github.com/paragonie/constant_time_encoding). (PHP didn't ship base32 either, so people who use this library get that too.)

Not shipping more encoding functions is probably the right choice.

jedisct1:
That RFC predates the web; I'm not sure that it is still relevant today.

If we have to implement one, picking one that doesn't require URL encoding would probably be a better choice.

BurningEnlightenment:
> That RFC predates the web; I'm not sure that it is still relevant today.

I always thought that [RFC4648](https://tools.ietf.org/html/rfc4648) would be _the_ base64 standard :confused: and at least the creator/maintainer of a _very_ fast base64 codec implementation thinks like this, too: https://github.com/aklomp/base64/issues/27#issuecomment-255472737 (I know that this is not representative in any way, but your debate reminded me of it)

> Still, having great, constant-implementations of these would be very useful, but I think this would be a project on its own.

I second this; base64 is a data encoding and not a crypto algorithm which therefore doesn't belong to a crypto library, but to some external codec library.

Just my 2ct.

PR #525, opened 2017-04-04
Fix ref10 slide on PPC-VLE

cfogelklou:
I have added some intermediate variables to slide() which removes whatever ambiguity was causing the crash with optimization flags on PPC_VLE architecture.  An added bonus is that, at least on PPC, the resultant assembly code is somewhat smaller, and probably faster.

I have forced the project to build using REF10 by removing ASM defines from the Makefile after ./configure, and it passes all 65 tests (at least on OSX)


jedisct1:
Merged, thanks!

Issue #533, opened 2017-04-21
crypto_pwhash_str_verify does not distinguish errors


emilbayes:
`crypto_pwhash_str_verify` does not distinguish between hash mismatch and the OS refusing to allocate memory. The underlying `argon2_verify` however does. It should be fairly simple to expose.

I will do a PR in one of the coming days to fix this :)

emilbayes:
@jedisct1 Would the best method here be a `errno` set for mismatch or a new return code?

jedisct1:
Hi Emil,

And thanks a lot for reporting this.

`errno` would be best!

PR #541, opened 2017-05-07
Add crypto_pwhash_MISMATCH errno

emilbayes:
This will attempt at #533. I do not know however if setting `errno` is considered the best way, or returning another error code would be better. As I see it `-1` is the expected "mismatch" error code and `ENOMEM` was an "unexpected" error cause from `crypto_pwhash_str_verify`, so if returning a error code is better we should decide on a error code for this.

### DIFF
--- a/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c
+++ b/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c
@@ -196,18 +196,9 @@ crypto_pwhash_argon2i_str_verify(const char str[crypto_pwhash_argon2i_STRBYTES],
         return -1;
     }
     /* LCOV_EXCL_STOP */
-    switch (argon2i_verify(str, passwd, (size_t) passwdlen)) {
-      case ARGON2_VERIFY_MISMATCH:
-        errno = EINVAL;
-        return -1;
-
-      case ARGON2_MEMORY_ALLOCATION_ERROR:
-        errno = ENOMEM;
-        return -1;
-
-      default:
-        errno = ENOMSG;
-        return -1;
+    if (argon2i_verify(str, passwd, (size_t) passwdlen) == ARGON2_VERIFY_MISMATCH) {
+      errno = EINVAL;
+      return -1;
     }
 
     return 0;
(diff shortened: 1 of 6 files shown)


jedisct1:
Setting `errno` to user-defined values doesn't sound great. It can conflict with existing codes, and I'm not even sure that `errno` is guaranteed to be an `int`.

Maybe set it to `EINVAL` instead?

emilbayes:
EINVAL seems to fit semantically but in this case it would make the cause of the error ambiguous. Is there another error that fits the semantics and isn't used by the verify function?

BurningEnlightenment:
> I'm not even sure that `errno` is guaranteed to be an `int`.

It is guaranteed to be of type `int` (see http://man7.org/linux/man-pages/man3/errno.3.html)

jedisct1:
`EINVAL` looks like the closest we can use, especially since it's defined everywhere, including Windows.

emilbayes:
@jedisct1 Then the error will be ambiguous given these lines: https://github.com/emilbayes/libsodium/blob/5b47dc8187312b326aa1b9f640433099cf3a6bba/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c#L199-L203

emilbayes:
@jedisct1 I understand what you're saying now. Please review my latest commit. I decided to use a `switch` to distinguish between lack of memory and password mismatch. Is that ok?

jedisct1:
`ARGON2_VERIFY_MISMATCH` should set `errno`, but in other cases, I'm not sure we need to change `errno`, as the calls to `malloc()`, `mmap()` or `posix_memalign()` will have already set it to the actual error code.

Also `ENOMSG` is a POSIX extension, so it is not available everywhere. I would just change the `switch` statement to `if` check against `ARGON2_VERIFY_MISMATCH`.

emilbayes:
Alright, learning lots here!

jedisct1:
The tests are failing, probably due to some `&& errno != EINVAL` that should actually be `|| errno != EINVAL`.

jedisct1:
This is now introducing a nice vulnerability. `crypto_pwhash_argon2i_str_verify` should return `-1` for any error, not just `ARGON2_VERIFY_MISMATCH`.

emilbayes:
Oh! Should have stuck with `switch`

PR #546, opened 2017-05-19
[.NET Core] allow win7 targetting

BurningEnlightenment:
From my point of view there is no reason to make this package win10 exclusive or am I blatantly missing something @ektrah?


jedisct1:
Comments, @ektrah ?

Issue #504, opened 2017-03-05
Pre-release for .NET Core


ektrah:
This issue is intended to track the progress towards a pre-release of a libsodium package for .NET Core.

The purpose of the package is to provide pre-compiled binaries for a number of platforms for the implementation of .NET Core language bindings such as [nsec](https://github.com/ektrah/nsec). The package does not provide a .NET API itself.

**Issues**

- [x] Tooling

  Add the scripts and files needed to build a libsodium package for .NET Core to the libsodium repository.

- [ ] Package metadata

  Metadata includes the package name, description and URLs. See the [libsodium.props](https://github.com/jedisct1/libsodium/blob/master/packaging/dotnet-core/libsodium.props) file. To avoid confusion, the description should IMO indicate that the package is intended for use with .NET Core only and not usable stand-alone.

- [x] Versioning

  ~~.NET Core uses [semver](http://semver.org/) for versioning: `major.minor.patch`. This means libsodium version numbers can be used as-is, but each new package release (e.g., when adding support for a new platform or if a package contains a broken binary) would always require a new libsodium version. This could be avoided, for example, by using different version numbers for the packages or switching libsodium to `major.minor` version numbers.~~

  Although .NET Core uses [semver](http://semver.org/) for versioning, it is possible to use version numbers with four parts for packages: `major.minor.patch.build`. This means that `major.minor.patch` can be used for the libsodium version, and `build` to number the packages released for a libsodium version.

- [ ] Logo

  Optional, but nice to have. Should be 128x128 and should look good when resized to 64x64 and 48x48.

- [ ] Building the package

  See the [README.md](https://github.com/jedisct1/libsodium/blob/master/packaging/dotnet-core/README.md) file.

- [ ] Publishing the package

  This requires an account at [nuget.org](https://www.nuget.org/). It's possible to grant other people uploading rights for a package once created, so if you like, I could take over the building and publishing steps.

BurningEnlightenment:
I would like to drive this effort further forward. I'm currently augmenting the package files to also support the AnyCPU .net framework target which will (eventually) enable [libsodium-net](https://github.com/adamcaudill/libsodium-net) to also use this package instead of distributing their own binaries.
[...]

PR #547, opened 2017-05-19
[.NET nuget] support .NET framework >= 4.6

BurningEnlightenment:
Add MSBuild targets for net46 target framework which copy both x86 and
x64 libsodium binaries into corresponding folders. This allows AnyCPU
applications to load the appropriate binary at runtime via
`LoadLibraryEx`.

The solution is inspired by what the SQLitePCL.raw people are deploying and should be sufficient for most use cases. Known exceptions are:
* Shadow Copying (Haven't looked too deeply into this, but it looks like there is no general solution)
* Mono (actually I haven't tested this, but I'm pretty sure there is some additional stuff necessary)

/cc @ektrah 
/xref #504

### DIFF
--- a/packaging/dotnet-core/desktop.targets
+++ b/packaging/dotnet-core/desktop.targets
@@ -1,0 +1,16 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+    <ItemGroup>
+        <None Include="$(MSBuildThisFileDirectory)..\..\runtimes\win7-x86\native\libsodium.dll">
+            <Link>x86\libsodium.dll</Link>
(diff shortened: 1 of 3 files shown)


jedisct1:
Merged, thanks!

ektrah:
I would really prefer to release the package for .NET Core first and add support for .NET Framework later, after a lot of testing and making sure that it works in a broad set of environments. The package works beautifully for .NET Core on Windows, Linux and macOS right now. With this pull request, it works only in a very narrow set of cases for .NET Framework and requires nasty, platform-specific hacks such as LoadLibraryEx to get it even barely working. I don't like this being added now.

PR #553, opened 2017-05-27
Fixed project file where one too many lines was removed from project files

LammyL:
Fixed project file where one too many lines was removed from project file in change 87473fd causing the projects to not build.


jedisct1:
Thanks!

PR #566, opened 2017-07-11
Update packaging for .NET Core

ektrah:
* Updates scripts to support package version numbers with four parts
* Adds `make test` to perform a quick test on the generated package
* Removes support for alpine.3-x64
* Adds support for fedora.26-x64


jedisct1:
Thanks!

PR #574, opened 2017-08-01
Build libsodium for i386 architecture

albertolalama:
Creates a universal binary for OSX architectures x86_64 and i386.


jedisct1:
The last OSX version that could be installed on 32-bit CPUs was Snow Leopard. That's really old, and Apple completely stopped supporting it for years now.

Xcode can still compile 32-bit binaries but it wouldn't be suprising to see that being removed without any prior notice. And even if the compiler itself still supports this, all the frameworks have been 64-bit only for a long time.

So we can't realistically support this. It's easier to assume that all OSX systems are 64 bit.

PR #576, opened 2017-08-02
[Findsodium.cmake] DON'T OVERWRITE LIBRARY SUFFIXES

BurningEnlightenment:
This fixes a _very_ nasty and spurious bug in the cmake find script 😓

Bug Description:
When using the find script for finding msvc shared libraries all following find_library() calls won't find any library, because they will be looking for files with the wrong suffix.


jedisct1:
Thanks!

PR #581, opened 2017-08-13
Update packaging for .NET Core

ektrah:
* Removes support for osx.10.10-x64, osx.10.11-x64, centos.7-x64, fedora.24-x64,  opensuse.42.1-x64, and ubuntu.16.10-x64 as they are no longer supported
* Replaces support for debian.8-x64, fedora.25-x64, fedora.26-x64, ubuntu.14.04-x64, and ubuntu.16.04-x64 with support for any Linux x64 distribution that has a sufficiently recent GLIBC

(See: [.NET Core 2.0 - Supported OS versions](https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md))

The presence of the .NET Framework 4.6 files in the package added by @BurningEnlightenment  still cause problems with the current .NET Core 2.0 pre-release. This pull request does not fix that yet. If the problems are still present with the final .NET Core 2.0 version, we may have to remove .NET Framework support.

@jedisct1: It would be awesome if you could include the NuGet package in your regular release dance, so that new libsodium versions are timely available (even when there are no changes to the packaging scripts). Command for the next pre-release: `python3 prepare.py 1.0.13-preview-01 && make`. (Please pull updated docker images first.)

### DIFF
--- a/packaging/dotnet-core/prepare.py
+++ b/packaging/dotnet-core/prepare.py
@@ -17,8 +17,6 @@ MACOS = [
   # --------------------- ----------------- #
   # Runtime ID            Codename          #
   # --------------------- ----------------- #
-  ( 'osx.10.10-x64',      'yosemite'        ),
-  ( 'osx.10.11-x64',      'el_capitan'      ),
   ( 'osx.10.12-x64',      'sierra'          ),
   # --------------------- ----------------- #
 ]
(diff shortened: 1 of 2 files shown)


jedisct1:
Thank you!

BurningEnlightenment:
@ektrah It's been a while since I looked at your packaging scripts, but IIRC we shouldn't need the other linux and mac recipes anymore, do we?

ektrah:
@BurningEnlightenment Most recipes are indeed not used right now, but it doesn't hurt to keep them. Some may actually be useful later again. E.g., there is work underway to bring .NET Core to Alpine Linux, which isn't based on GLIBC.

jedisct1:
Just wanted to say a big *thank you* to both of you for your invaluable help regarding .NET core!

PR #582, opened 2017-08-15
Update packaging scripts to .NET Core 2.0

ektrah:
This updates the packaging scripts to .NET Core 2.0 and fixes two issues:

* When packing the libsodium package, `dotnet pack` using .NET Core 2.0 gives the following error:

    > error MSB3644: The reference assemblies for framework ".NETFramework,Version=v4.6" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed. Note that assemblies will be resolved from the Global Assembly Cache (GAC) and will be used in place of reference assemblies. Therefore your assembly may not be correctly targeted for the framework you intend.

    The fix is to remove `net46` from `<TargetFrameworks>`. It doesn't seems to be strictly needed anyway.

* When adding the libsodium package to a .NET Core 2.0 project, `dotnet add package libsodium --version 1.0.12-preview-01` gives the following warning:

    > warning NU1701: Package 'libsodium 1.0.12-preview-01' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project.

    I've tried a few things to get rid of the warning, but so far I haven't found anything working. @BurningEnlightenment  - Do you have an idea?

    The fix is to remove support for .NET Framework for now.

----

The libsodium NuGet package [is now a dependency](https://github.com/tabrath/libsodium-core/pull/10) of the [libsodium-core](https://github.com/tabrath/libsodium-core) project.

@tabrath @rwasef1830 - The libsodium NuGet package is intended for projects that wrap libsodium in a .NET library, like yours. The goal is to provide a stable, high-quality, well-tested NuGet package containing the native assets, primarily for .NET Core.

The package is currently marked as "preview" because there are currently some issues that need to be addressed. This may require some changes to the package before it can be marked as stable.

If you want, you can help speed up things by testing the package in as many configurations as possible. In particular, there hasn't been much testing related to ASP.NET yet (shadow copying, etc.).


rwasef1830:
@ektrah I already tested it on a net461 targeting ASP.net Core project through Sodium.Core 1.0.5 and it works, libsodium.dll is copied to output and works fine.

BurningEnlightenment:
@ektrah I'll take a look at it this evening.

jedisct1:
That's pretty awesome!

BurningEnlightenment:
@rwasef1830 I'm very curious how you got it to work on net46(1), because on my machine the build of net46(1) targets is just broken, because _nothing_ is copied.
As to the dotnet 2.0 failure: I think I read something about the problem and how the dotnet core team worked around it; hopefully we can find a way to support both, the full framework and the core 2.0 standard.
***
And by the way, I still don't get why @tabrath duplicated the dotnet core port of libsodium and didn't even try to merge the changes back into upstream 😕

tabrath:
@BurningEnlightenment I'm nothing against that, actually. But when I started working on it, it diverged so much from your code I thought it might be too much? Would you consider a merge of the projects?

rwasef1830:
@BurningEnlightenment 
This compiles correctly and copies the 32-bit version of libsodium.dll to output (because <PlatformTarget> defaults to x86):

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net461</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="libsodium" Version="1.0.12-preview-01" />
  </ItemGroup>

</Project>
```

BurningEnlightenment:
@rwasef1830 Yes, that's pretty much what I've done there: https://github.com/BurningEnlightenment/libsodium-net/blob/netcore/libsodium-net/libsodium-net.csproj
And that works very well with the current upstream nuget package, but at least not for me with a nuget built with the updated packaging script, i.e. a nuget package missing the targets.desktop 😕 
***
> I'm nothing against that, actually. But when I started working on it, it diverged so much from your code I thought it might be too much? Would you consider a merge of the projects?

@tabrath I would love to get the libsodium-net project reunified and I really hope that we can merge efforts, but I won't be able to put any effort into it before friday. I'll open an issue on your project as soon as I get to it, so we don't completely derail this thread; sorry for the outlash 😓

tabrath:
@BurningEnlightenment Sounds good to me! Looking forward to it 😄

rwasef1830:
@BurningEnlightenment I tested only on the current publicly available nuget package, not including any changes caused by this pull request.

I am not highly familiar with the intricacies of NuGet beyond the basics, what's the reason a package like this one even has a special net461 build ? Wouldn't it be enough to target netstandard1.0 and get universal coverage, since it is just the binaries packaged as Nuget .. (no actual .net code of any kind).

ektrah:
@rwasef1830 - There are differences how .NET Core and .NET Framework locate native DLLs. The .NET Core story is very smooth; you just include the DLLs in the directory in the NuGet package. The .NET Framework story is much more messy; you need to write scripts that hook into MSBuild and copy the DLLs into the directory of each project that uses the DLLs. This causes a lot of trouble related to platform bitness (x86, x64), ASP.NET shadow copying, and now .NET Core 2.0 compatibility.

rwasef1830:
@ektrah I am a bit confused here, when you make an exe project as I mentioned in the snippet above, the latest msbuild seems to take care of this automatically, I see libsodium.dll in the output directory beside the .net exe. Msbuild even copies the right bitness version depending on ```<PlatformTarget>``` value (which default to x86).

The problem you described does occur when you reference Libsodium from a DLL (example: Unit test project), msbuild doesn't do that magic for class library projects, but a simple workaround for unit tests is simply to make it an EXE with NUnitLite or similar (though that may not be always possible I admit).

For transitive reference, the MSbuild automatic handling works... ie: Asp.net core project -> C# Library -> Libsodium, I still get libsodium.dll in the Asp.net core project output and it is loaded and works fine.

Is there an example of the issue you described ? (Forgive my ignorance of the matter, I haven't delved into this repository and are not familiar with what's happening here).

ektrah:
@rwasef1830 - MSBuild automatically copying native DLLs is news to me. Can you point me to some documentation or examples, or provide some steps that I could reproduce locally?

tabrath:
I think the issue is with older versions of msbuild, right? When I try using mono (net461), it's failing.

rwasef1830:
@ektrah It is not well documented unfortunately, I was struggling with this myself for a while. This post is old but documents the concept (some details changed in newer msbuild): https://blog.3d-logic.com/2015/11/10/using-native-libraries-in-asp-net-5/ 

Then I resorted to dissecting the libuv package and looking asp.net core source code: https://github.com/aspnet/libuv-package/blob/dev/src/Libuv/project.json (translate it the the new csproj by hand).

An example of this I did in a pull request before I removed it and started depending on Libsodium package. Just make a new blank VS2017 style netstandard1.0 csproject file, and add native libraries in runtimes folder exactly as libuv package (libsodium now already does this).

A working example of msbuild copying native Dlls is above in my comment to @BurningEnlightenment make a new empty solution, and add the csproj above and compile it and look in the bin/Debug folder, you will see libsodium.dll that comes from the NuGet package.

Note: There is a different bug in MSBuild related to copying, the native dll copying only works for PackageReferences, see my comment here: https://github.com/tabrath/libsodium-core/issues/5#issuecomment-322129752

Ie: A project referencing another with "runtimes" and right csproj tags will still not get native dll copying, the runtimes must be coming from a NuGet package (PackageReference).

BurningEnlightenment:
the copying for native runtimes on net461 only works if you are using the new project format and set your build target to a specific platform target, i.e. even if you have the new project format and build your project with anycpu and without the prefer 32-bit switch nothing is copied.

ektrah:
I've baked a fresh new libsodium NuGet package ([libsodium.1.0.13-justtesting.nupkg.zip](https://github.com/jedisct1/libsodium/files/1226269/libsodium.1.0.13-justtesting.nupkg.zip); remove the .zip extension), and referenced that from a .NET Framework 4.6.1 console application in VS 2017.3. The libsodium.dll is not found. What am I missing? It works beautifully in a .NET Core application.

ektrah:
It indeed does work when setting the platform target to x86 or x64 (but not when set to AnyCPU)!
@rwasef1830  @BurningEnlightenment @tabrath  - What do you think - is that "good enough"?

tabrath:
@ektrah Sounds reasonable, doesn't it? What runtime should be used if you don't specify any?

rwasef1830:
@ektrah I'm more puzzled, with the 1.0.13 it doesn't work for me :-( I'll have to do some testing to narrow this down.

rwasef1830:
@ektrah Interesting package: I wonder if libsodium referenced that, would all problems get solved ? https://www.nuget.org/packages/microsoft.netcore.platforms/

ektrah:
@rwasef1830 - Could you try these?

* [libsodium.1.0.13-preview-01-test-a.nupkg.zip](https://github.com/jedisct1/libsodium/files/1231716/libsodium.1.0.13-preview-01-test-a.nupkg.zip)
* [libsodium.1.0.13-preview-01-test-b.nupkg.zip](https://github.com/jedisct1/libsodium/files/1231717/libsodium.1.0.13-preview-01-test-b.nupkg.zip)
* [libsodium.1.0.13-preview-01-test-c.nupkg.zip](https://github.com/jedisct1/libsodium/files/1231711/libsodium.1.0.13-preview-01-test-c.nupkg.zip)

*(Remove the .zip extension.)*

All of them work in a .NET Framework 4.6.1 x64 console application project in Visual Studio 2017 version 15.3 on my machine.

rwasef1830:
Results:
```
- netstandard1.1 dll - Libuv: No
- netcoreapp1.1 exe - Libuv: No
- netcoreapp1.1 exe x86 - Libuv: No
- net461 dll - Libuv: No
- net461 exe - Libuv: No
- net461 exe x86 - Libuv: Yes

- netstandard1.1 dll - 1.0.12-preview-01: No
- netstandard1.1 dll - 1.0.13-justtesting: No
- netstandard1.1 dll - 1.0.13-preview-01-test-a: No 
- netstandard1.1 dll - 1.0.13-preview-01-test-b: No
- netstandard1.1 dll - 1.0.13-preview-01-test-c: No
- netstandard1.1 dll - 1.0.13-preview-01-test-d: No

- netcoreapp1.1 exe - 1.0.12-preview-01: No
- netcoreapp1.1 exe - 1.0.13-justtesting: No
- netcoreapp1.1 exe - 1.0.13-preview-01-test-a: No
- netcoreapp1.1 exe - 1.0.13-preview-01-test-b: No
- netcoreapp1.1 exe - 1.0.13-preview-01-test-c: No
- netcoreapp1.1 exe - 1.0.13-preview-01-test-d: No

- netcoreapp1.1 exe x86 - 1.0.12-preview-01: No
- netcoreapp1.1 exe x86 - 1.0.13-justtesting: No
- netcoreapp1.1 exe x86 - 1.0.13-preview-01-test-a: No
- netcoreapp1.1 exe x86 - 1.0.13-preview-01-test-b: No
- netcoreapp1.1 exe x86 - 1.0.13-preview-01-test-c: No
- netcoreapp1.1 exe x86 - 1.0.13-preview-01-test-d: No

- net461 dll - 1.0.12-preview-01: Yes (x86 and x64 folders)
- net461 dll - 1.0.13-justtesting: No
- net461 dll - 1.0.13-preview-01-test-a: No
- net461 dll - 1.0.13-preview-01-test-b: No (dumped whole lot unrelated .NETStandardLibrary DLLs)
- net461 dll - 1.0.13-preview-01-test-c: No
- net461 dll - 1.0.13-preview-01-test-d: No

- net461 exe - 1.0.12-preview-01: Yes (x86 and x64 folders)
- net461 exe - 1.0.13-justtesting: No
- net461 exe - 1.0.13-preview-01-test-a: No
- net461 exe - 1.0.13-preview-01-test-b: No (dumped whole lot unrelated .NETStandardLibrary DLLs)
- net461 exe - 1.0.13-preview-01-test-c: No
- net461 exe - 1.0.13-preview-01-test-d: No

- net461 exe x86 - 1.0.12-preview-01: Yes (in addition to x86 and x64 folders)
- net461 exe x86 - 1.0.13-justtesting: Yes
- net461 exe x86 - 1.0.13-preview-01-test-a: No
- net461 exe x86 - 1.0.13-preview-01-test-b: Yes (and dumped whole lot unrelated .NETStandardLibrary DLLs)
- net461 exe x86 - 1.0.13-preview-01-test-c: Yes
- net461 exe x86 - 1.0.13-preview-01-test-d: No
```

Sample project used for testing:
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>VALUE</TargetFramework>
    <OutputType>VALUE</OutputType>
    <PlatformTarget>VALUE</PlatformTarget>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="libsodium" Version="VALUE" />
  </ItemGroup>

</Project>
```
I changed the 4 values by hand for each combination and called ```dotnet restore && dotnet clean && dotnet build``` looking at the correct bin\Debug folder after. Got lazy to write a script :-D

dotnet --version: 1.1.0

Visual Studio 2017 15.3 is installed but remained closed throughout all cases.

Hope that helps. It seems I was wrong about the platforms package :-(
Also, I find it strange no dll is copied at all in case of netcoreapp1.1, but I do see libsodium referenced in a .deps.json file generated beside the output dll.

rwasef1830:
Related old bug: https://github.com/aspnet/KestrelHttpServer/issues/1292

ektrah:
@rwasef1830 - Thanks for the many tests! I was actually only interested in the last ones (net461 exe x86).

Regarding netcoreapp1.1: As I mentioned above, .NET Core locates native DLLs differently from .NET Framework. The DLLs no not need to copied for .NET Core; they are taken from the extracted NuGet package (C:\Users\\<user\>\\.nuget\\packages\\libsodium\\<version\>\\ on my machine). Only for .NET Framework the DLLs must be copied.

I've reproduced your tests for net461 exe x86 on my machine. Could you check if 1.0.13-justtesting really copies the DLL? Apart from the package name, it is exactly identical to 1.0.13-preview-01-test-a and does *not* copy on my machine.

test-b and test-c are different from the others in that they have a non-empty dependency group. test-b references NETStandard.Library 1.6.1, test-c references Microsoft.NETCore.Platforms 1.0.1.

I seems we can make everyone happy with test-c. (We don't want all those DLLs from NETStandard.Library.)

ektrah:
BTW, there is a section on the NU1701 warning in the [.NET Standard 2.0 announcement](https://blogs.msdn.microsoft.com/dotnet/2017/08/14/announcing-net-standard-2-0/).

rwasef1830:
@ektrah I confirm your findings, -justtesting doesn't copy, I made a mistake in the table. Edited to fix.

LammyL:
If I understand the original problem here, we had to remove net461 from the project because it won't compile the package on linux, but if we remove net461 from the nuget package, the dll's don't copy the dll's over if the framework project is set to Any CPU.  

If this is right, I managed to get this to work, I think, so that it will copy the dll's over and successfully pass the tests for libsodium.net against .net framework.

I started by comparing the differences between the contents of the 1.0.12-Preview-01.nupkg and 1.0.13-Preview-01.nupkg.  The only differences i found were these: **build\net46\libsodium.targets** file and **<group targetFramework=".NETFramework4.6" />** in **libsodium.nuspec** file.

So I took the libsodium-1.0.15.nupkg file, made those two changes, then repacked the nuget package with **nuget pack libsodium.nuspec**.  Then I uploaded it to my local nuget repository and installed it into the libsodium-net project.  Then I ran the tests and everything worked.

If you get the same results that I do, I'm wondering if we can tweak the build process to fix this.
- Build normally on linux like we do right now for netstandard
- Then unzip the nupkg file and make the necessary changes
- Then repack the nupkg file and publish it to nuget.org.

*also note the paths in the 1.0.12-Preview-01 version of build\net46\libsodium.targets needed to be tweaked to win-x64 instead of win7-x64.

LammyL:
The other thing I think we should do in the build project is explicitly declare netstandard1.3 AND netstandard2.0 as target frameworks, because Visual Studio 2017 will incorrectly choose the net461 framework instead of netstandard1.3 framework if you are referencing the nuget package from a netstandard2.0 library.

ektrah:
The problem is patching the build process with the .targets file. I think it's amazing that the package works for .NET Framework *without* the .targets file if you set the platform to x86 or x64. Trying to make it work for AnyCPU for .NET Framework is not worth the hassle. (AnyCPU for .NET Core works fine.) Just compile for x64 (or use .NET Core) and be happy :-)

PR #584, opened 2017-08-24
WIP: Meson build system

whs:
I wanted to learn [Meson](https://github.com/mesonbuild/meson), and I'm already using libsodium in my app so I decided to experiment with porting libsodium to Meson.

This is a work in progress. I wanted to see what the maintainers think about it before trying to finish it.

## Why switch to Meson?
Also see [why systemd moved to Meson](https://in.waw.pl/~zbyszek/blog/systemd-meson.html)

### Faster build
I did a test on two machines:

- A: 2015 MBP, 2.7GHz i5, 8GB RAM, Flash storage running macOS, building inside a Docker container (xhyve driver) running debian:stretch.
- B: i5-6500 3.2GHz, 8GB RAM, SSD running up-to-date Arch Linux as of 2017-08-24

Here's some not-exactly-scientific benchmark:

Command | Machine A | Machine B
---------------|----------|---------|
./autogen.sh | 7.398s | 6.005s
./configure --with-pthreads --enable-static=no --enable-opt | 10.372s | 5.437s
make -j5 | 1m12.168s | 25.298s
make check -j5 | not tested | 3.473s
make install | 0.368s | 0.310s
**Autotools time (no autogen)** | 82.908s | 40.523s
--- | ----
meson -Doptimize=true -Dpthread=true --buildtype release | 4.556s | 2.813s
ninja  | 13.591s | 12.119s
nina test | not tested | 1.664s
ninja install | 0.258s | 0.372s
**Meson time** | 18.405s | 16.968s

Note:

- Machine A result is currently outdated
- Meson also build and link test executables, therefore `ninja test` does not include compilation time. To compensate, `make check` is timed using its second run where everything should be already compiled.
- Some tests are disabled in Meson as they're broken

## Less complexity
Some features of autoconf were replaced with Meson's

- --enable-debug were removed in favor of Meson's [--buildtype](http://mesonbuild.com/Running-Meson.html#configuring-the-source)
- --enable-valgrind were removed in favor of Meson's [test helper](http://mesonbuild.com/Unit-tests.html#testing-tool)
- PIC support is provided directly by Meson and is no longer controlled by our build system.
- [**Meson does not support GCC older than 4.7**](http://mesonbuild.com/Design-rationale.html) The check to enable flax-vector-conversions for gcc older than 4.3 were removed.
- pkgconfig are now generated by [Meson module](http://mesonbuild.com/Pkgconfig-module.html) instead of using the template.
- CWFLAGS support is removed. Just set c_args directly.
- NDEBUG is no longer `undef`-ed as Meson doesn't read environment variable. Meson option `b_ndebug` can be used to enable NDEBUG (default off).
- **Only shared OR static library are built at a time**. Use `--default-library` to switch and rebuild. (https://github.com/mesonbuild/meson/issues/484)
- ~--disable-pie might be removed in favor of Meson's `b_staticpic`. Is both the same?~
- [x] Some warnings might be able to enable by turning on Meson's warning_level.

Also, Meson's build file is a lot clearer to understand than Autotools'.

Finally, Meson can generate Visual Studio solution file so we don't need to ship them separately.

## Status
- [x] Builds on Linux (tested with Arch and debian:stretch Docker)
- [x] Builds on Mac
- [ ] Builds on Windows
- [ ] Builds on NaCl
- [ ] Builds on PNaCl
- [ ] Builds on Emscripten (https://github.com/mesonbuild/meson/issues/1802)
- [ ] Builds on Android
- [ ] Read config from generated config file instead of using compiler defines (I might need help implementing this in Autoconf)
- [ ] Unit testing
- [ ] What to do with Safecode? Is this replaceable with something in Meson?
- [ ] Update CI tool configurations
- [x] Implement AX_CHECK_CATCHABLE_SEGV
- [x] Implement __STDC_LIMIT_MACROS check
- [x] ~Data alignment check. Can we use `compiler.alignment()` to check?~ `CPU_UNALIGNED_ACCESS` is never used
- [ ] msys/mingw/cygwin detection
- [ ] Doesn't compile on clang due to intrin.h detection (https://github.com/mesonbuild/meson/issues/2246) -- If the check is manually overriden the build works just fine
- [ ] Check whether it is possible to check for linker flag support in Meson (currently it is implemented as `-Wl,--xxx` compiler argument check which seems to always return true)
- [ ] Add meson files to autotools' EXTRA_DIST

One possible way to merge this faster is to ship with Autotools for some tricky situations (NaCl/PNaCl/Emscripten/MSYS/Mingw/Cygwin), and Meson for standard builds (Win/Lin/Mac. This would replace MSVC solution files).


jedisct1:
Maybe a different build system will be used for libsodium 2.x or for the final version of libhydrogen.

But for libsodium 1.x, what we have now perfectly works, doesn't require Python, does what we want and has been battle tested for years on many environments. Language bindings, packages and projects embedding libsodium heavily depend on it.

Rewriting exactly what we already have with a different tool can only introduce regressions. So, we're going to keep using what already works and doesn't have any problems to be solved that would require rewriting everything for another tool.

FSMaxB:
@whs: If you were to maintain meson support out of tree that would be quite useful, because it would make using libsodium as a meson subproject much easier. Maybe even with an entry in wrapdb.

whs:
I'll revise the scope to only support win/lin/mac and submit it to WrapDB.

Right now the only things left are to test with Windows, and make sure all tests work.

Issue #588, opened 2017-08-30
Question about Integer Multiplication (Ref10 Curve25519)


paragonie-scott:
There are a few places in the ref10 code where an integer is multiplied by `1 << n` instead of being bitshifted by `n`. Is there a reason for this? Bonus: Would replacing these operations with a left shift make the code faster?

https://github.com/jedisct1/libsodium/blob/6b43c1ddb6bb7d7f2528bdac8433520603150345/src/libsodium/crypto_core/curve25519/ref10/curve25519_ref10.c#L240-L270

https://github.com/jedisct1/libsodium/blob/6b43c1ddb6bb7d7f2528bdac8433520603150345/src/libsodium/crypto_core/curve25519/ref10/curve25519_ref10.c#L351-L421

Related to my inquiry, but not any specific security issues:

* https://research.kudelskisecurity.com/2017/01/16/when-constant-time-source-may-not-save-you/
* https://bearssl.org/ctmul.html

paragonie-scott:
(Sending a pull request, mostly to see if Travis CI pukes if I make this change.)

jedisct1:
These are signed integers.

When negative, shifting a signed integer has an undefined behavior in C.

paragonie-scott:
Ah. This is why I don't do C development. :)

Thanks, @jedisct1 :)

jedisct1:
Carries could be unsigned. But everything else has to be signed.

PR #589, opened 2017-08-30
Change some multiplications to equivalent shifts

paragonie-scott:
See #588 for background.


jedisct1:
In practice, the generated code is exactly the same. But left shifting a negative value is undefined in C.

PR #605, opened 2017-09-29
Save libsodium.js

buu700:
See https://github.com/jedisct1/libsodium.js/pull/106.


jedisct1:
This is pretty awesome!

Trying this immediately.

buu700:
Nice, hope it works! I mentioned this in the other PR, but I haven't tested this at all (so there could be random typos in any of the JS or bash, etc.), and I may have missed some references when changing/deleting things; so there's almost definitely gonna be some additional work needed on top of what I've done, but hopefully this is a good enough starting point that it won't be much more than what I would've had to ask for your help with anyway.

dist-build/emscripten.sh
```diff
@@ -62,9 +63,32 @@ emmake make clean
...
+  emccLibsodium "${PREFIX}/lib/libsodium.asm.tmp.js" -Oz -s RUNNING_JS_OPTS=1 -s NO_EXIT_RUNTIME=1
+  emccLibsodium "${PREFIX}/lib/libsodium.wasm.tmp.js" -O3 -s WASM=1
+
+  cat > "${PREFIX}/lib/libsodium.js" <<- EOM
+    if (typeof Module === 'undefined') {
+      var Module = {};
+    }
+    Module.ready = new Promise(function (resolve, reject) {
+      Module.onAbort = reject;
+      Module.onRuntimeInitialized = function () { resolve(); };
```
buu700: Oops, this should just be `Module.onRuntimeInitialized = resolve;`. Originally it was running `resolve(Module)` before I changed a few things, but now it's slightly unnecessarily verbose. Won't submit a new PR just for that now that it's been merged, but just pointing it out for your reference @jedisct1.

jedisct1:
Combining the asmjs and wasm in one file greatly simplifies things. But also forces users to download twice as big as what they actually need, and the library isn't exactly small already.
Since applications have to wait for the promise to be ready anyway, maybe keeping both versions separated and load only the required one could still be a good idea.

buu700:
Hm, well, maybe there could be one build that has them combined and one that doesn't? (For my use case at least I'd need the latter.) It's possible that it won't actually be _that_ much bigger as-is, though, due to the increased compactness of both wasm and fully minified `-Oz` asm.js.

PR #610, opened 2017-10-05
libsodium.js totalMemory minor fix

buu700:
Something small I noticed while looking at the latest commit.


jedisct1:
Good catch!

PR #621, opened 2017-10-27
CMake find module fixes for static linking:

mellery451:
Remove option() declaration from module since it hides the normal variable in
parent scope (thuys set() in parent does not work). Hardcode static lib
name for find_library since pkgconfig does not provide it.

### DIFF
--- a/contrib/Findsodium.cmake
+++ b/contrib/Findsodium.cmake
@@ -32,7 +32,9 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU"
 endif()
 
 # static library option
-option(sodium_USE_STATIC_LIBS "enable to statically link against sodium")
+if (NOT DEFINED sodium_USE_STATIC_LIBS)
+    option(sodium_USE_STATIC_LIBS "enable to statically link against sodium" OFF)
+endif()
 if(NOT (sodium_USE_STATIC_LIBS EQUAL sodium_USE_STATIC_LIBS_LAST))
     unset(sodium_LIBRARY CACHE)
     unset(sodium_LIBRARY_DEBUG CACHE)
@@ -53,6 +55,11 @@ if (UNIX)
     endif()
 
     if(sodium_USE_STATIC_LIBS)
+        # if pkgconfig for libsodium doesn't provide
+        # static lib info, then override PKG_STATIC here..
+        if (NOT DEFINED sodium_PKG_STATIC_LIBRARIES)
+            set(sodium_PKG_STATIC_LIBRARIES libsodium.a)
+        endif()
         set(XPREFIX sodium_PKG_STATIC)
     else()
         set(XPREFIX sodium_PKG)


contrib/Findsodium.cmake
```diff
@@ -32,7 +32,6 @@ if (CMAKE_C_COMPILER_ID STREQUAL "GNU"
 endif()
 
 # static library option
-option(sodium_USE_STATIC_LIBS "enable to statically link against sodium")
```
BurningEnlightenment: instead of removing this, I would very much prefer something like this:
```cmake
set(sodium_USE_STATIC_LIBS OFF CACHE BOOL "enable to statically link against sodium")
```
which can be overridden by just setting up the cached variable before issuing a call to `find_package(sodium)`. See the [docs](https://cmake.org/cmake/help/v3.10/command/set.html#set-cache-entry) for reference

contrib/Findsodium.cmake
```diff
@@ -53,6 +52,9 @@ if (UNIX)
     endif()
 
     if(sodium_USE_STATIC_LIBS)
+        # pkgconfig for libsodium doesn't provide
+        # static lib info, so override PKG_STATIC here..
+        set(sodium_PKG_STATIC_LIBRARIES libsodium.a)
```
BurningEnlightenment: I would like to see this guarded just in case this changes in the future:
```cmake
if (NOT DEFINED sodium_PKG_STATIC_LIBRARIES)
    ...
endif()
```

mellery451: good suggestion - I will add that.

contrib/Findsodium.cmake
mellery451: From the calling side, I want to be able to do this:

```
set(sodium_USE_STATIC_LIBS ON)
find_package(sodium REQUIRED)
```

This doesn't work if you "declare" the variable inside the find module using `option()` or with `set( .... CACHE ...)`.  Just looking around at the various Find* modules provided with CMake, you'll see that most of them *don't* try to declare the variable -- probably because it prevents the simple usage I've outlined above. Yes, if we declare it a CACHE variable in the module, I could override it in my CMakeLists, but I would have to do:

```
set(sodium_USE_STATIC_LIBS ON CACHE BOOL "why are you making me do this?")
```

So, for that reason, I advise we simply check the variable and not try to declare it as a cache/option - again, that is what most standard Find* modules do.

BurningEnlightenment: The rationale behind the switch was that the precompiled binaries for windows always provide shared and static versions and it really is a matter of preference whether or not to use static or shared libraries, so from my point of view it made sense to provide a visible option.

However, I do understand that there are projects for which this agility isn't wanted/necessary and I want to support those too, so I think guarding the option with an if not defined clause is the most sensible approach.

Lastly I recommend squashing all those changes into one commit in order to reduce the noise created by these small changes.

mellery451: ok - I think the defined check around option works for me. Squashed and pushed - let me know if it looks right to you.

jedisct1:
Merged! Thanks that both of you!

nibua-r:
Maybe I missed something but it doesn't seem to work for me.
I have made a simple project to describe what happens by my side: https://github.com/nibua-r/sodium-issue-621
Feedbacks are welcome!

BurningEnlightenment:
@nibua-r thanks for the heads up; I will take a look tomorrow.

PR #624, opened 2017-10-30
Resolved static analyser issues

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. Warnings, found using PVS-Studio:
src/libsodium/crypto_stream/salsa20/xmm6int/u1.h  15      warn    V760 Two identical blocks of text were found. The second block begins from line 87.
src/libsodium/crypto_stream/salsa20/xmm6int/u0.h  14      warn    V760 Two identical blocks of text were found. The second block begins from line 86.

So, we can add lamda-like function or put it into cycle `for`.

### DIFF
--- a/src/libsodium/crypto_stream/salsa20/xmm6int/u0.h
+++ b/src/libsodium/crypto_stream/salsa20/xmm6int/u0.h
@@ -7,153 +7,83 @@ if (bytes > 0) {
     __m128i b0, b1, b2, b3, b4, b5, b6, b7;
     uint8_t partialblock[64];
 
-    unsigned int i;
+    unsigned int i, j;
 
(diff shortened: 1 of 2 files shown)


jedisct1:
What bug does it solve? Does it improve performance?

jedisct1:
You don't need to introduce another variable. If you remove half of the loop, just increment `i` by `2` instead of `4`.

Trying to force it to be unrolled was probably not a bad idea, especially when the library is not compiled with `--enable-opt`. Actual benchmarks on recent CPUs would be required to confirm this.
If you can perform these, that would be awesome.

In the meantime, this pull request cannot be merged, if only because the introduction of another variable doesn't make any sense.

Issue #625, opened 2017-10-31
detecting lack on entropy sources


chantra:
libsodium version: libsodium-1.0.15

`sodium_init` aborts when no source of entropy is found. In https://www.nlnetlabs.nl/bugs-script/show_bug.cgi?id=2141 unbound would abort without any user facing warning because the user process chroot into a directory where no source of entropy is defined.
The lack of error message makes it harder to troubleshoot or to give the user some actionable error.

This can be reproduced with the following program:
https://gist.github.com/chantra/4c13f696b3645961d6749651d1a093f7

I originally expected that the RC code of `sodium_init` would be enough to decide whether or not the system would pass the checks, but apparently not.
What I am trying to find is a portable way to detect that the initialization will fail in order to give an actionable item to the end user. I though of using `sodium_set_misuse_handler` but this does not seem to work. Here is how I used it: https://gist.github.com/chantra/df4f94992aea0109bd21d6646d4de208

Am I not using `sodium_set_misuse_handler` properly? should the handler be called before aborting? If not, what is a good way to detect whether or not the entropy sources are available other than stabbing at /dev/random (unless it it sufficient to do so?).

Note: I remember reading that one way to sort this is to call `sodium_init` before  chroot-ing so a handle to /dev/random is held open. While that may be doable, I would rather avoid the initialization of dnscrypt earlier in unbound initialization.

jedisct1:
This should be fixed in the `stable` branch.

But the best way do handle chroot environments is as you described, to call `sodium_init()` before the `chroot` call. You don't need to initialize anything else. It's also safe to call `sodium_init()` again later if you have to.

chantra:
ok, so I tried against master, still no avail. the issue is that when entering `sodium_init`, the lock is held: https://github.com/jedisct1/libsodium/blob/802830e4e6c1f3fb3d00c79e8e2c99b029d3efa6/src/libsodium/sodium/core.c#L43

Then we enter `randombytes_stir`: https://github.com/jedisct1/libsodium/blob/802830e4e6c1f3fb3d00c79e8e2c99b029d3efa6/src/libsodium/sodium/core.c#L53

which fails:
https://github.com/jedisct1/libsodium/blob/802830e4e6c1f3fb3d00c79e8e2c99b029d3efa6/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c#L267

which in turn requires a lock:
[...]

PR #626, opened 2017-11-01
[freebsd] use arc4random on FreeBSD

chantra:
This would help while running a process in a chroot without /dev/random
on FreeBSD, just like on OpenBSD
See #625

Using https://gist.github.com/chantra/910305a456cbcf2a7fc8aeaa259036fe
and assuming using libsodium-stable

https://github.com/jedisct1/libsodium/commit/5d484e6cb2ea1f2d4d052e31627be6ee2e2477c3

before we would get:
```
root@freebsd-512mb-sfo2-01:~ # env
LD_PRELOAD=/tmp/lsodium/lib/libsodium.so ./sodium_init
sodium_set_misuse_handler
Misused handler!
```

after:
```
root@freebsd-512mb-sfo2-01:~ # env
LD_PRELOAD=/tmp/lsodium/lib/libsodium.so ./sodium_init
sodium_set_misuse_handler
Hello world,
```

### DIFF
--- a/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c
+++ b/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c
@@ -55,7 +55,7 @@ BOOLEAN NTAPI RtlGenRandom(PVOID RandomBuffer, ULONG RandomBufferLength);
 # pragma comment(lib, "advapi32.lib")
 #endif
 
-#if defined(__OpenBSD__) || defined(__CloudABI__)
+#if defined(__OpenBSD__) || defined(__FreeBSD__) || defined(__CloudABI__)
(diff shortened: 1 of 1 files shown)


jedisct1:
Unlike OpenBSD, FreeBSD's `arc4random()` still uses RC4.

Until a couple months ago, it could also silently fall back to using `time()`+`getpid()` as a seed.

So, we'd better stick with `/dev/urandom` in this platform for now.

chantra:
Oh OK, very fair point :)
The misuse handler will be good enough to at least prompt the user to fix the chroot.
Thanks!

lattera:
FreeBSD, starting with 12-CURRENT, uses chacha20 instead of RC4. I can't remember off-hand if the change made it to 11.1-RELEASE.

PR #629, opened 2017-11-13
Add crypto_pwhash_BYTES const

emilbayes:
I think it would be good to have a `crypto_pwhash_BYTES` like there is a `crypto_generichash_BYTES`. I know that the "easy path" should be to use `crypto_pwhash_str`, but I think it would be good for user experience to have the bytes for symmetry with other APIs, and it doesn't add much maintenance overhead.

### DIFF
--- a/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c
+++ b/src/libsodium/crypto_pwhash/argon2/pwhash_argon2i.c
@@ -25,6 +25,12 @@ crypto_pwhash_argon2i_alg_argon2i13(void)
 }
 
 size_t
+crypto_pwhash_argon2i_bytes(void)
+{
+    return crypto_pwhash_argon2i_BYTES;
+}
+
+size_t
 crypto_pwhash_argon2i_bytes_min(void)
 {
     COMPILER_ASSERT(crypto_pwhash_argon2i_BYTES_MIN >= ARGON2_MIN_OUTLEN);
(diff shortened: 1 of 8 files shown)


jedisct1:
Hi Emil,

I'm not sure it makes sense to have such a constant defined, what its value would be and that it wouldn't add more confusion than be useful.

This function can be considered as a XOF. The output can be of any length, because it was not tailored to a specific use case or other algorithm. For example, it can be used to derive multiple keys of different sizes at once.

The limits (`MIN` and `MAX`) are obviously needed. But some default value... not any more than having one for `randombytes_buf()`.

emilbayes:
I think I may be looking at the functions with too narrow a context. I was asked what the default length of the hash for storing passwords should be, if you were not to use the "easy" `crypto_pwhash_str` functions, and did not find a recommended length like many of the other primitives. But I also think you are right in that this might lead others to the conclusion that this API is not a general purpose KDF, but only for password storage.

jedisct1:
Yes, even if they obviously rely on the same underlying primitive, these functions are designed to serve different purposes, and documenting them this way avoids confusion.

`pwhash_str()` should really be the go-to API for password storage, if only because it also stores the algorithm identifier along with all other parameters required to recompute the hash. Something frequently overlooked when using the raw API.

emilbayes:
So I think the conclusion here is that it is better that I make a PR towards the documentation, rather than adding another constant, which will narrow other users understanding of the primitive.

PR #642, opened 2017-12-04
[CMAKE] Fix static library detection on unix

BurningEnlightenment:
Manually prepend `lib` and append `.a` to the static library names provided by the cmake pkg-config cmake module. This is necessary because the `find_library()` function always prefers the shared object file if the library name is "unqualified" and unfortunately the cmake pkg-config module seems to provide only "unqualified" library names through `*_PKG_STATIC_LIBIRARIES` (which makes this part of the cmake pkg config module rather useless for most cases 😒).

After inspecting the pkg config module source I also found out that the pkg config variables are _always_ set and therefore never undefined which is why I replaced the not defined case.

@mellery451 and @nibua-r can you test this, please?
@jedisct1 please do not merge yet, thanks :wink:

/xref #621

### DIFF
--- a/contrib/Findsodium.cmake
+++ b/contrib/Findsodium.cmake
@@ -55,23 +55,35 @@ if (UNIX)
     endif()
 
     if(sodium_USE_STATIC_LIBS)
+        foreach(_libname ${sodium_PKG_STATIC_LIBRARIES})
+            if (NOT _libname MATCHES "^lib.*\\.a$") # ignore strings already ending with .a
+                list(INSERT sodium_PKG_STATIC_LIBRARIES 0 "lib${_libname}.a")
+            endif()
+        endforeach()
+        list(REMOVE_DUPLICATES sodium_PKG_STATIC_LIBRARIES)
+
         # if pkgconfig for libsodium doesn't provide
         # static lib info, then override PKG_STATIC here..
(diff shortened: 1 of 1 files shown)


jedisct1:
Got it :)

contrib/Findsodium.cmake
```diff
@@ -55,23 +55,35 @@ if (UNIX)
...
         # static lib info, then override PKG_STATIC here..
-        if (NOT DEFINED sodium_PKG_STATIC_LIBRARIES)
+        if (sodium_PKG_STATIC_LIBRARIES STREQUAL "")
             set(sodium_PKG_STATIC_LIBRARIES libsodium.a)
         endif()
+
         set(XPREFIX sodium_PKG_STATIC)
     else()
+        if (sodium_PKG_LIBRARIES STREQUAL "")
+            set(sodium_PKG_IBRARIES sodium)
```
nibua-r: Should be set(sodium_PKG_**L**IBRARIES sodium)

BurningEnlightenment: fixed 🙈

jedisct1:
Merged! Thanks guys!

nibua-r:
Thanks for this great lib!!! 👍

PR #653, opened 2017-12-27
Updates for latest emscripten

buu700:
As of the latest version of emscripten available via the sdk (1.37.27):

* `-s NO_DYNAMIC_EXECUTION=1` is basically a no-op (https://github.com/kripken/emscripten/issues/5911)

* `--closure 1` can be safely enabled for better minification due to the above

* `-s NO_EXIT_RUNTIME=1` is a no-op (https://github.com/kripken/emscripten/pull/5878)


jedisct1:
Thank you!

jedisct1:
Using the closure compiler only saves 207 bytes. Not really significant, and it can only introduce bugs.

buu700:
Ah okay, fair enough.

PR #666, opened 2018-01-19
Fix: undefined SIZE_MAX when building android

KangLin:
Fix: undefined SIZE_MAX when building android


jedisct1:
Isn't `<stdint.h>` more appropriate?

jedisct1:
Added in 729d80a1accabc24eb0d8b26a487fef3ea43bc6c

Thanks for pointing this out!

KangLin:
There are not <stdint.h> when android version less then 21,

PR #670, opened 2018-01-23
Include current fuzzers, test them as part of travis

derwolfe:
I'm working towards trying to get an [ideal integration](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md) for the fuzzers that were added as part of https://github.com/google/oss-fuzz/pull/1051.

This PR includes the fuzzers and _hopefully_ a way that they will build successfully on travis. I'll add the caveat that I've not worked with make for a long time so it might be gut-wrenchingly bad :D.

The PR contains:
1. A standalone fuzz target that can be used to execute a single fuzz input. This is used to verify that the fuzzers can run and don't bit rot.
2. Some semi-outdated seed corpora for the fuzzers currently in oss-fuzz. If this seems reasonable, I can go grab the most recently minimized seeds and include them instead.

If/when this is ok for merge, I can get the fuzzers that are currently living in OSS-fuzz removed and alter the build script to use the ones saved here.

Thanks :)


.travis.yml
```diff
@@ -26,6 +26,7 @@ script:
  - ./configure --disable-dependency-tracking --enable-minimal
  - make check
  - ( echo '#include <sodium.h>' ; echo 'int main(void) { return sodium_init(); }' ) > /tmp/main.c && gcc -Isrc/libsodium/include -Isrc/libsodium/include/sodium $(find src -name '*.c' -o -name '*.S') /tmp/main.c
+ - cd fuzz && make fuzz-check
```
derwolfe: It might be simpler/more straightforward just to make another row in the build matrix that runs the fuzz-check rule.

jedisct1:
Great work!

However, especially since this is a work in progress, maybe this should be in a separate repository that Travis would check out before running the tests?

Here's the current test:

```c
  setup_fake_random(data, size);
  unsigned char key[crypto_auth_KEYBYTES];
  unsigned char mac[crypto_auth_BYTES];

  crypto_auth_keygen(key);
  crypto_auth(mac, data, size, key);
  crypto_auth_verify(mac, data, size, key);
```

Fuzzing is useful to find inputs that will violate assertions.

Here, we don't have any assertions at all, except "the process doesn't exit prematurely".
Static analyzers such as Tis can prove that there are no indirect memory accesses that depend on `data` and `key`, so if one input doesn't trigger an out-of-bounds memory access, all of them will behave the same. This also helps to prove that the code runs in constant time for any given secrets.

Fuzzing may not even catch out-of-bound accesses here, since we're just accessing unguarded memory on the stack. Sodium has `sodium_malloc()` that can be used to immediately crash on out-of-bounds accesses, including reads.

Besides this, what could we be looking for? What other assertions do we want our fuzzer to violate?

We have a function that generates a key, and one that uses it.
An assertion is that any output of the `keygen()` function is a valid key input for the `auth()` function.
If we find a case where `auth()` fails for a key produced by `keygen()`, we found a bug. `fails` might be that it returns a non-zero error code, that it doesn't change the output buffer, or that it leaks the key in that output buffer.

We have a `verify()` function. An assertion is that given the same parameters as `auth()`, the `verify()` function must always return `0`. If we find an input to `auth()` that doesn't satisfy this, we found a bug.

We also have equivalences. `crypto_auth_hmacsha512256()` and `crypto_auth()` must produce the same output given the same input. If they don't, we found a bug. That case is not very interesting and would be caught with the regular test suite, but there are more interesting equivalences to look for.

We have a streaming API, `512256_init()`, `512256_update()`, `512256_final()`. The assertion is that `final(update(m0), update(m1), update(m2), ... update(mn))` produces the same result as `h(m0 || m1 || m2 ... || mn)`, no matter what `n` is, and the individual length of the message parts.
If we find a set of messages that produce a different hash than hashing a single pre-concatenated message, we found a bug.

Fuzzers are amazing at finding cases like these. But fuzzers need some help to know what assertions stand, and what they should be looking for, beyond application crashes.

derwolfe:
Thank you very much for your detailed feedback! :)

Neither of the current fuzzers seem like they do much outside of checking for crashers, which given what you've said above seems like a highly unlikely result. 

What do you think of the following steps?

1. Close this PR.
2. Open a PR to oss-fuzz to disable the libsodium build until newer, more useful fuzzers are in place.
3. Work on writing one or more the fuzzers you described (listed under Possible New Fuzzing Targets for posterity).

Would you mind if I submit PRs for new fuzzers to this repo under the proposed `/fuzz` directory? I'd be happy to leave the exhaustive/larger corpora seeds out at the start and use a simple single test for each, but I would like to have building the fuzzers coupled with libsodium's build process. This would have two benefits (1) the fuzzers are regularly built as part of CI and (2) the fuzzers that make it in will be of higher quality with your reviews :).

# Possible New Fuzzing Targets

> If we find a case where auth() fails for a key produced by keygen(), we found a bug. fails might be that it returns a non-zero error code, that it doesn't change the output buffer, or that it leaks the key in that output buffer.

> We have a verify() function. An assertion is that given the same parameters as auth(), the verify() function must always return 0. If we find an input to auth() that doesn't satisfy this, we found a bug.

> We have a streaming API, 512256_init(), 512256_update(), 512256_final(). The assertion is that final(update(m0), update(m1), update(m2), ... update(mn)) produces the same result as h(m0 || m1 || m2 ... || mn), no matter what n is, and the individual length of the message parts. If we find a set of messages that produce a different hash than hashing a single pre-concatenated message, we found a bug.

jedisct1:
Hi Chris,

Sounds like a good plan. But I would still use a separate repository, if only because the fuzzer project will have its own history; and you will have the liberty to rebase and merge things until they are stable if you want to. It will be cleaner.

Travis will check the fuzzer in a `fuzz` directory, so it will be part of every CI build.

derwolfe:
Sorry, for whatever reason I mentally jumped over pulling the repo in for CI. Thanks :)

PR #671, opened 2018-01-23
add configure option to use size_t for length params instead of unsigned long long

mvduin:
The API's bizarre use of unsigned long long instead of size_t for lengths has bothered me for a long time. This patch set adds a configure option to use size_t instead, and a typedef to allow code to work with both variants.

Tested to pass "make check" both with and without --enable-size_t on x86_64-pc-linux-gnu and arm-linux-gnueabihf.


jedisct1:
Sorry, we can't break the ABI at this point.

But as you probably noticed, libhydrogen uses `size_t` for sizes, so this is what libsodium 2.0 will use as well.

mvduin:
There is no ABI break unless you deliberately pass the --enable-size_t option. By default it remains ABI compatible.

jedisct1:
This is dangerous. The same version of the same library can have two different ABIs. For bindings and dynamically linked applications, this is a disaster waiting to happen.

And inventing a new type is worse than using `unsigned long long`.

Sorry, no.

mvduin:
The only part I agree on is the risk in case of dynamic linking, although this is solvable by making --enable-size_t imply --disable-shared. If there is an interest in a shared version (I personally don't care), it also wouldn't be hard to give the library name some suffix when --enable-size_t is used.

And a typedef is not ideal, but it is the only way to smoothly transition from unsigned long long to size_t: even if unsigned long long remains the default for backwards-compatibility, introducing this type would allow end-user code to start becoming agnostic of the choice, hence be forward-compatible with a future where size_t is used.

jedisct1:
I also hate the `unsigned long long` thing. But introducing such a hack is too dangerous.

If you want to be forward-compatible, you should use the libhydrogen API.

mvduin:
I very much disagree this is a "hack", and the only danger identified is avoided if --enable-size_t is made to inhibit building shared libraries. Nothing changes in the default configuration anyway, but this patch does allow people to use libsodium without infecting their own code with this long-long bullshit, as long as they're okay with statically linking against a non-default variant of libsodium.

libhydrogen is neither compatible with nor interoperable with libsodium, hence not relevant here.

PR #675, opened 2018-02-07
README: link to libsodium-doc repository

enkore:
Previously the libsodium-doc repo didn't seem to be mentioned anywhere (README, finished docs), which made it a bit non-obvious where the docs actually come from


jedisct1:
Thanks again!

PR #682, opened 2018-02-18
Add spectre v2 migitations for GCC

loganaden:
GCC 7.3 has support for retpoline. I see that the previous patch was reverted due to the use of -mretpoline. retpoline support is enabled via -mfunction-return=thunk and -mindirect-branch=thunk, in GCC.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -202,6 +202,9 @@ AC_CHECK_DEFINE([_FORTIFY_SOURCE], [], [
 AX_CHECK_COMPILE_FLAG([-fvisibility=hidden],
   [CFLAGS="$CFLAGS -fvisibility=hidden"])
 
+AX_CHECK_COMPILE_FLAG([-mfunction-return=thunk -mindirect-branch=thunk],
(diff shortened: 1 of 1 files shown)


loganaden:
Also, tested with make check, and here are the results:

Testsuite summary for libsodium 1.0.16
============================================================================
# TOTAL: 72
# PASS:  72
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0

jedisct1:
This was reverted after discussing with the LLVM people working on SPECTRE mitigations.

The kernel, and anything running untrusted code should apply these mitigations. However, in the context of libsodium, even called via bindings running untrusted PHP code, it doesn't appear to be useful.

Can you describe an attack scenario that this would protect against?

loganaden:
I agree with you. In OpenSSH, there was a similar discussion where a rogue process running on a host could target it and cause it to leak its secrets, using spectre variation 2. Patching   libsodium would eliminate this attack. (https://lists.mindrot.org/pipermail/openssh-unix-dev/2018-February/036656.html)

Issue #685, opened 2018-02-21
libsodium in SGX


rbost:
For a project, I would like to compile libsodium so it can be used in an Intel SGX enclave. This would (from what I see and understand at this moment) require changes to the autoconf script and some additional code for the RNG.

I will do that myself, so as to learn more about libsodium and SGX, but I wanted to know if that was something that would be considered for inclusion in libsodium's source when it will be done?

jedisct1:
That would definitely be awesome!

I was planning to do it for libhydrogen, but if you feel like doing it for libsodium, go, go, go!

rbost:
So, I have a first working version [here](https://github.com/rbost/libsodium/tree/sgx).

Using the resulting compiled lib, I was able to build an app hashing the same string in the enclave and out of the enclave. And you know what? They hashed to the same value!!!!

For the moment, I see one security issue: sgx_read_rand generates pseudo-random bytes (and not truly random ones) when building the enclave in simulation mode, and I don't think this is good at all. I believe that, even in simulation mode, libsodium must remain cryptographically sound. Yet I don't not see a secure straightforward solution. Ideas are appreciated ...

The question of how to test the build remains: it would be great if we could test the execution of the library inside SGX by building an app that would create an enclave that would run the tests. Yet, that would require changing a lot the testing architecture that relies on many independent executables each testing a single feature. Again, suggestions will be necessary 😃

rbost:
This is exactly the conclusion I came to, and I am not super thrilled about any of those two options.

Strictly speaking, I think that you can use the simulation mode on a platform that does not support RDRAND. So, if you want to build something that is still cryptographically correct, this would be a problem (sure, one might ask why we want to do this in the simulation mode, and what is the point in term of security, but still, we shouldn't jeopardise the crypto of the library in any way).
Also, if you look at the [implementation of `sgx_read_rand`](https://github.com/intel/linux-sgx/blob/master/common/src/sgx_read_rand.cpp), it seems that whenever RDRAND is available, it is used (even in SIM mode). It is also where you realize that, despite what is announced, there is a way to call CPUID from within the enclave without making an OCALL. It is annoying, but it is another problem ...

[...]

PR #688, opened 2018-03-04
Port libsodium to SGX

rbost:
PR corresponding to Issue #685.
I got it working on an Ubuntu 16 LTS VM in simulation mode (I do not have access to an Intel chip with SGX enabled, and could not test it in hardware mode). 

I was not able to figure out where to add the documentation related to the changes, but once I am told, I would be happy to make the changes.


rbost:
@Moschn Sorry about that, I left a typo in the autoconf file ... (see commit b85df91).

This being fixed, I was able to run ./autogen.sh &&  ./configure --enable-sgx && make -j && make check without any issue. It would be great if you could test the code again :)

Moschn:
@rbost I tried it and indeed it compiles now. But all tests fail with the error `Illegal Instruction`. Some lines from the output:
```
../../build-aux/test-driver: line 107: 19285 Illegal instruction     (core dumped) "$@" > $log_file 2>&1
FAIL: aead_aes256gcm
../../build-aux/test-driver: line 107: 19292 Illegal instruction     (core dumped) "$@" > $log_file 2>&1
FAIL: aead_chacha20poly1305
../../build-aux/test-driver: line 107: 19299 Illegal instruction     (core dumped) "$@" > $log_file 2>&1
FAIL: aead_xchacha20poly1305
```
I can also provide you with the core dumps if you need them. 

Intel SGX disallows [some instructions](https://software.intel.com/en-us/node/703005) inside an enclave. I think it might be possible to disable some of these instructions with `./configure` but I have not experimented with this.

rbost:
@Moschn This is great news: somethings happens!

It would be great if you could send me the core dumps.

I knew about the forbidden instructions, and genuinely thought I did not use them. My hope is that this is coming from the SGX-only code I wrote for the tests. If it comes from the inners of libsodium, this is going to be a lot less fun ...

Moschn:
Sent it to you via email

rbost:
@Moschn Unfortunately, I could not use the core dumps you sent me in a very useful way. However, after some thinking, and staring at the code, I realized that I did not remove the calls to sodium_malloc (and sodium_free) in the test code, which would create SIGILL (on a syscall). So I did change that in 41d65f88370fc171ccf7134d02c4e46ad08fe04f.

Also, in 9f3d485859f796428b7a153caa3d044ebd125551, I added some more logging to exactly see more precisely where the code fails without relying on a debugger.

Again, some testing from your side would be great!

Moschn:
@rbost I still get the same errors:
```
../../build-aux/test-driver: line 107:  9944 Illegal instruction     (core dumped) "$@" > $log_file 2>&1
FAIL: aead_aes256gcm
```

rbost:
@Moschn :( I will take a last look at the code to see what happens, but I am not sure I will find anything else. And I don't want to bother you too much.
So unless someone else debugs it, I will have to get my hands on a machine with SGX to get further.
Thanks a lot for your help anyway!!

jedisct1:
This is probably not going to happen.

I don't have the required hardware to test this, this looks complicated to install on Linux and impossible elsewhere, I don't have any use case for this, this branch is unmaintained, CI cannot be used to test this either. If there is any interest in this, it should better be maintained independently.

PR #689, opened 2018-03-05
Store emscripten-symbols as JSON

webmaster128:
This is the first step on the way to generate wrapper, documentation and TypeScript definitions depending on the standard/sumo configuration. This converts emscripten-symbols.def into JSON format, in order to make it easy to process from different languages.

This adds an additional python3 dependency for the person executing generate-emscripten-symbols.sh, but I guess this script is only run by maintainers.

The idea was to make this first step as small as possible, so the long inline python command probably disappears soon. I guess most of generate-emscripten-symbols.sh can easily be converted into Python, which improves readability.

In order to improve readability and reduce diff size, [Process Substitution](http://mywiki.wooledge.org/ProcessSubstitution) is used, which requires bash. Is this an issue?

By the way, why do wee need to look at the symbols in /usr/local/lib/libsodium.23.dylib? This makes the script unusable on Linux. Could we move this into a separate, optional verification call?


jedisct1:
I'm not sure that such changes should be made here. Having them in libsodium.js, that already requires quite a few dependencies, would make more sense.

The definition file is updated manually, and having it as a JSON file makes it less readable.

All the symbols defined here are also not guaranteed to be present in the resulting library. So, even if it is a bit more cumbersome, checking if symbols are actually present in libsodium.js is more robust.

It is currently done only at runtime, but it can be done when generating the wrapper as well.

webmaster128:
Thanks for the feedback!

I was thinking about where this issue belongs and tried to do it in libsodium.js. The main reason I did it here was to avoid any kind of duplication to avoid files being out of sync.

I'll give inspecting libsodium.js a try

PR #691, opened 2018-03-06
Run-time wasm test before resolving ready

buu700:
Just to clarify exactly what I was suggesting [here](https://github.com/jedisct1/libsodium.js/issues/139#issuecomment-370675994). Haven't tested, but I think it'll fix the problem.


jedisct1:
It should!

Thanks, Ryan!

PR #692, opened 2018-03-06
Adding option to enable a full build

kalaspuffar:
We are working on a solution to use scrypt in libsodium-jni.

I'm trying to work this out together with @joshjdevl.

I understand if this is not the correct solution but it's a suggestion from my side give us the option to build the full libsodium.


jedisct1:
Merged, thanks!

PR #693, opened 2018-03-07
using POSIX getconf for number of available processors

joshjdevl:
Attempt to detect the number of processors for concurrent make


jedisct1:
Merged, thanks!

I get `getconf: _NPROCESSORS_ONLN: unknown variable` so we may want to hide `stderr` though.

PR #699, opened 2018-03-17
add crypto_scalarmult_ed25519_noclamp

hf:
`crypto_scalarmult_ed25519(q, n, p)` clamps the scalar `n` to a suitable
value for finding the point on the curve. This is limiting in scenarios
where the scalar multiplication should be done over unclamped scalars
such as field elements arising out of field arithmetic. An example of
such a use case is the Schnorr NIZK protocol from RFC8235.

### DIFF
--- a/src/libsodium/crypto_scalarmult/ed25519/ref10/scalarmult_ed25519_ref10.c
+++ b/src/libsodium/crypto_scalarmult/ed25519/ref10/scalarmult_ed25519_ref10.c
@@ -28,9 +28,9 @@ _crypto_scalarmult_ed25519_clamp(unsigned char k[32])
     k[31] |= 64;
 }
 
-int
-crypto_scalarmult_ed25519(unsigned char *q, const unsigned char *n,
-                          const unsigned char *p)
+static inline int
+_crypto_scalarmult_ed25519(unsigned char *q, const unsigned char *n,
+                          const unsigned char *p, const unsigned int clamped)
 {
     unsigned char *t = q;
     ge25519_p3     Q;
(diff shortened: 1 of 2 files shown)


hf:
**How it came to be?**

I was trying to implement RFC8235. After two excruciating days in trying to find why modular arithmetic suddenly stopped working in the world, I decided to remove all clamps from libsodium. Modular arithmetic started working again.

I think this should go into libsodium as it's the only Curve25519 implementation I can portably use in my applications, and as well includes a lot of other good stuff like Argon2, Blake2B, etc.

**Tests**

I was not sure how to write a test for this case, I'm open to suggestions.

tarcieri:
Naively removing clamping potentially opens you up to small subgroup attacks arising from the cofactor.

RFC 8235 says:

> Let G be a base point on the curve that serves as a generator for the subgroup over E(Fp) of prime order n.  The cofactor of the subgroup is denoted h, which is usually a small value (not more than 4). 

Ed25519 is cofactor 8. I don't think this is going to work.

hf:
@tarcieri that is true, but I would not read the RFC as a specification of which ECs you should use, or their parameters.

The RFC is simply explaining how a Schnorr NIZK protocol would be implemented regardless of which EC you use. 

I'm open to suggestions on how to rename the functions so that it's clear to users that this is not a function you want to use normally, but only in specific implementations of protocols. Similar to the `crypto_core_ed25519_add` and `crypto_core_ed25519_sub` functions.

hf:
Interestingly libhydrogen already does this: https://github.com/jedisct1/libhydrogen/blob/master/impl/x25519.h

jedisct1:
"not a function you want to use normally, but only in specific implementations of protocols" is out of the scope of libsodium.

tarcieri:
> @tarcieri that is true, but I would not read the RFC as a specification of which ECs you should use, or their parameters.
>
> The RFC is simply explaining how a Schnorr NIZK protocol would be implemented regardless of which EC you use.

If they're including language like "not more than 4" for the cofactor, that should give you pause to consider whether a cofactor 8 curve is an appropriate choice. I believe for the particular construction described in this RFC, the answer is: no. Perhaps you should consider asking the [curves mailing list](https://moderncrypto.org/mailman/listinfo/curves).

hf:
@tarcieri I will write there, although I think I came up with a solution without clamping (but with the same cofactor)

PR #712, opened 2018-04-29
Use _MESSAGEBYTES_MAX in crypto_aead_xchacha20poly1305

tom-auger:
I noticed that `crypto_aead_xchacha20poly1305_ietf_MESSAGEBYTES_MAX` was defined but not actually used. 

This will change the message length check in `crypto_aead_xchacha20poly1305_ietf_encrypt` to compare against:
```
UINT_MAX - crypto_aead_xchacha20poly1305_ietf_ABYTES
```
on 32 bit platforms, rather than:
```
UINT64_MAX - crypto_aead_xchacha20poly1305_ietf_ABYTES
```


jedisct1:
Thanks!

PR #714, opened 2018-05-11
Fix CPU feature detection crash on VS 2010 32-bit /O2

etodd:
This is a crash in a very specific build configuration, where the optimizer put the cpu_features pointer in eax, and then eax got blown away by the feature detection asm. Turning off optimizations for this function solved the problem.

### DIFF
--- a/src/libsodium/sodium/runtime.c
+++ b/src/libsodium/sodium/runtime.c
@@ -109,9 +109,6 @@ _cpuid(unsigned int cpu_info[4U], const unsigned int cpu_info_type)
 #endif
 }
 
-#ifdef _MSC_VER
-#pragma optimize( "", off )
-#endif
 static int
 _sodium_runtime_intel_cpu_features(CPUFeatures * const cpu_features)
 {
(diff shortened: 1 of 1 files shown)


jedisct1:
That looks like a kludgy and unreliable workaround we can't rely on. 

We should rather push/pop `%eax` and `%ecx` in the assembly code.

jedisct1:
Maybe `%edx` as well.

jedisct1:
Quoting: https://msdn.microsoft.com/en-us/library/k1a8ss06.aspx

> When using __asm to write assembly language in C/C++ functions, you don't need to preserve the EAX, EBX, ECX, EDX, ESI, or EDI registers.

That doesn't match what you report. Or was that guarantee introduced after VS2010?

jedisct1:
Actually, Microsoft wrote the same thing about VS2010: https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/k1a8ss06(v=vs.100)

jedisct1:
Can you still try to save the GPRs, and see if that fixes the issue you are seeing?

etodd:
To be more clear, this is the specific #ifdef activated by this build configuration:

    # elif defined(_MSC_VER) && defined(_M_IX86)

jedisct1:
The registers being changed by `xgetbv` is expected, but the compiler should save and restore the ones holding values used both before and after the `__asm` block.

 So, either a compiler bug, or a different root cause.

Can you add

```asm
push eax
push ecx
push edx
```

to the beginning of the block, and

```
pop edx
pop ecx
pop eax
```

to the end?

etodd:
That also worked. Pretty weird stuff. Added those changes here.

jedisct1:
Added in cfb0f94704841f943a5a11d9e335da409c55d58a (with correct indentation and a comment to justify this).

Thanks!

PR #722, opened 2018-06-20
Fix for emscripten breaking change

buu700:
Not an immediate issue if libsodium is still stuck on an older version of emscripten pending resolution of https://github.com/kripken/emscripten/issues/6387, but this will be needed when you do upgrade to make sure that the asm.js fallback continues to work as intended.


jedisct1:
Thank you Ryan!

jedisct1:
But yes, we're still stuck to old Emscripten versions. https://github.com/kripken/emscripten/issues/6387 is still a thing.

PR #727, opened 2018-07-03
Better handling of --disable-asm option during build.

areading:
This changeset resolves an issue while building libsodium on non-amd64 platforms. Specifically, a build for x86 (i686 core, say) fails during compilation of certain SSE2 instructions in the xmm6int variant of salsa20 crypto_stream routines, even with the '--disable-asm' configure option. This is due to '--disable-asm' actually implying HAVE_AMD64_ASM=0; xmm6int SSE2 routines are included in the SSE2 library in that case.

These commits expose a new WITH_ASM automake variable for the '--disable-asm' option. WITH_ASM is tied to other asm-related build checks, so the existing HAVE_xxx_ASM configs will never be defined when ASM has been disabled (as expected). Also, '--disable-asm' now means no auxilliary library (sse,avx,aes-ni,etc.) will be built.

Two less-obvious behavioral changes I want to explicitly call out:
* Previously, CPUID check was executed if ASM was allowed OR '$host_alias' = 'x86_64-nacl'. These changes effectively turn this into an AND.
* Previously, the .private_extern directive checks were run if ASM was allowed OR '$host_os' = 'nacl'. Again, these changes effectively turn this into an AND.

Happy to discuss if anything seems amiss. Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,6 +88,9 @@ AS_IF([test "$host_os" = "nacl" -o "$host_os" = "pnacl"], [
   AC_MSG_WARN([compiling to Native Client - asm implementations disabled])
 ])
 
+AS_IF([test "x$enable_asm" = "xno"], [WITH_ASM=0], [WITH_ASM=1])
+AM_CONDITIONAL([WITH_ASM], [test "x$WITH_ASM" != "x0"])
+
 AC_ARG_ENABLE(pie,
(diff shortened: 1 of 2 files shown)


jedisct1:
I don't follow. The `xmm6int` implementation of Salsa20 doesn't use inline assembly.

And `--disable-asm` is only for NaCl and Emscripten targets.

What exact error do you get, on which platform and with which compiler?

PR #730, opened 2018-07-06
Add pwbox - a password secured secretbox

cleishm:
This is a proposal for a "password secured secret box". I'd love some feedback as to whether this would be a helpful addition, and whether it has been added in the most optimal way. For example, I was considering if it should be it's own box type (pwbox) or just additional methods for a secretbox (e.g. `crypto_secretbox_password`, `crypto_secretbox_password_open`, etc).

The intention of this construct is to simplify the common use-case of generating a secretbox key using pwhash and then creating a secretbox. It stores all pwhash parameters and the (randomly generated) secretbox nonce into the mac, allowing the box to be stored and later opened without persisting these separately.

Output includes a magic "pwb" prefix, to help guard against invalid inputs that could result in expensive pwhash calculations.


jedisct1:
Hi!

This is a useful thing and the design looks fine.

However, I don't think the library itself should be extended with new operations that can be achieved using simple compositions. Especially when they encourage the use of passwords instead of high-entropy keys.

While this solves one specific problem, its addition would be immediately followed by complaints that it is not suitable for file encryption. Or that it inherits the limitations of the current pwhash API (see [libhydrogen pwhash](https://github.com/jedisct1/libhydrogen/wiki/Password-hashing) for what it will eventually be replaced with).

Instead of directly providing these functions, these could be a fantastic example to add to the current documentation, similar to the [PHP example](https://github.com/jedisct1/libsodium-php#encrypt-a-file-using-a-key-derived-from-a-password).

cleishm:
So I did consider that, however, I looked at the way that password storage (e.g. `crypto_pwhash_str`) does have a construction in the library that embeds parameters into the output and thus produces results that are portable across implementations. Given the precedent, it seems reasonable to have a standard way for this parameter embedding in the library.

While the documentation does describe how to store parameters, it doesn't provide the certainty that different implementations embed parameters in the same way. Having the code in the library solves for this.

Can you elaborate a little more on your thoughts around this? I'm also curious about the limitations in the pwhash API you're concerned about?

jedisct1:
How do you encrypt a file? How do you change a password? How do you increase the work factor of the hash function? How do you mitigate offline attacks? How do you know if the password is correct without having to load and decrypt everything? How do you prevent a malicious/corrupted input from using very expensive parameters?

The current code doesn't address any of these. If it did, it would be overkill for some applications. And for other applications, there would still be missing features requiring a different format.

Which is why having this in the documentation or even as a complete tutorial might be better.

Or it can be an independent project, so you can get feedback from users about their needs and their thoughts on the API.

cleishm:
Right, this proposal doesn't address all those potential requirements. It does one specific thing in a way that would then be portable, much the same as password hash string construction is provided by the library.

I've seen this specific thing done a number of times, hence thought it might be nice to provide it generally. That would make it more easily usable, less error-prone, and portable - all attributes I believed this library was aiming for.

Obviously, I can implement it in a higher-level library, but that does compromise the goal of portability and common usability.

jedisct1:
Sounds like the actual missing piece in the puzzle is a way to serialize/unserialize `pwhash()` parameters. That could be a nice addition to libhydrogen as well.

cleishm:
Sure, why not. That is a bit broader in scope, however, as it suggests having the same serialization approach for all applications that embed the parameters (and seed/nonce) into the output.

If that is ok, I'd probably propose a format like:

```
<magic><alg><opslimit><memlimit>
```

Where opslimit and memlimit are 64bit longs _in network byte-order_ (big-endian).

cleishm:
I updated this PR to serialize those longs in BE, as an example for anyone who wants to add that suggested documentation, examples or (de)serialization method.

PR #741, opened 2018-08-02
Fix wrong statement about constant time sodium_memcmp()

ahamez:
The number of instructions depend on the parameter `len`.

### DIFF
--- a/src/libsodium/include/sodium/utils.h
+++ b/src/libsodium/include/sodium/utils.h
@@ -26,7 +26,7 @@ void sodium_stackzero(const size_t len);
 
 /*
  * WARNING: sodium_memcmp() must be used to verify if two secret keys
- * are equal, in constant time.
+ * are equal.
  * It returns 0 if the keys are equal, and -1 if they differ.
  * This function is not designed for lexicographical comparisons.
  */


angt:
yes...
but the important point is that it **only** depends on the parameter `len`.

ahamez:
Exactly, so the complexity is O(len), not O(1).

angt:
We are not talking about complexity here, we just want a constant time version of `memcmp`. Everyone will understand that the time obviously depends on the parameter `len`.

angt:
To be clear, `constant time` is the most important thing in the comment.

ahamez:
Sorry, but it's not clear to me then: what's performed in constant time here? Clearly, it's not the comparison as it takes `len` steps.

jedisct1:
Until we have massively parallel computers that can perform vector operations on vectors the size of the memory, CS 101 should be enough to make it obvious that loading data of arbitrary length is not what "constant time" can apply to.

jedisct1:
Removing the entire purpose of the function from its description can make more harm than good, as there won't be any apparent justification for using it over `memcmp()` any more.

`for a given length` could be added, but that would remain unnecessary overzealous pedantry.

jedisct1:
clowney

ahamez:
Yes, sorry for that :-) Thank you for the explanations!

ahamez:
For those like me that may end up here, here's an explanation of what is constant-time in the realm of cryptography: https://crypto.stackexchange.com/a/61263/56338.
So, @jedisct1, here's a suggestion: libsodium is intended for people like me who are not well versed into crypto, and that still want to secure their apps. Thus, maybe, it can be useful these define this property and why it is important, most certainly [here](https://download.libsodium.org/doc/helpers/).

jedisct1:
pr

angt:
There are places for PR, places for issues and places for questions.

Please note that libsodium is said "easy to use" but **I** would not recommend it for beginners that have no clue about sec.

PR #759, opened 2018-09-26
Update Findsodium.cmake

SlavSlavov:
When libsodium is build and installed from source, the line:
list(REMOVE_DUPLICATES sodium_PKG_STATIC_LIBRARIES)
generates an error because sodium_PKG_STATIC_LIBRARIES is empty
The proposed change fixes this issue


jedisct1:
Thanks!

PR #761, opened 2018-09-29
Provides explicit_memset supports/NetBSD.

devnexen:
Similar to explicit_bzero function is to defeat
compiler optimisation.


jedisct1:
Thanks!

PR #762, opened 2018-09-29
getrandom on FreeBSD support.

devnexen:
FreeBSD 12 has similar interface as Linuxes only
as a real function and with one mode which used here.


devnexen:
Only __FreeBSD_version is not a compiler's constant but a system's one thus sys/param.h inclusion is needed.

jedisct1:
Thanks!

This has been reverted until I figure out a workaround to avoid `tcc` segfaulting.

PR #763, opened 2018-09-29
useBackupModule getRandomValue fix

buu700:
Fixes a minor issue with #752 (run-time error when running any sodium method after `sodium.libsodium.useBackupModule()` has been called).


jedisct1:
👍

PR #770, opened 2018-10-09
Implement the draft-irtf-cfrg-vrf-03 VRF construction

algoradam:
This construction is specified in an IETF internet draft, https://datatracker.ietf.org/doc/draft-irtf-cfrg-vrf/03/ ; we implement the ECVRF-ED25519-SHA512-Elligator2 suite, which is similar to VXEdDSA


jedisct1:
We are not going to add something that is still a draft.

The chacha20poly1305 construction whose nonce size changed between two drafts is the worst thing that happened to this library, so let's not repeat this.

Also, VRFs are nice, but nobody ever asked for this. I have no immediate use for a VRF either, so this is not something I am willing to maintain, especially in libsodium.

The usual way to see if a new operation is worth adding is to offer it as a distinct project, then wait for feedback, see how people use it, if they use it at all, and finally see if it should be merged to the library or not.

PR #790, opened 2018-12-03
Made sig parameter of crypto_sign_final_verify() const

ivmaykov:
The functions `crypto_sign_final_verify()` and `crypto_sign_ed25519ph_final_verify()` only read the contents of the `sig` parameter, so it can be safely made `const`. This makes the API easier to use for callers who only have a const pointer to the signature (especially in C++, which requires a verbose `const_cast<unsigned char*>(sig)` currently).


jedisct1:
Indeed! Thank you!

PR #791, opened 2018-12-03
Add a public API to sign/verify pre-hashed messages with Ed25519ph

### LINKED ISSUE #789: Add a public API to sign/verify pre-hashed messages with Ed25519ph
I would like to use libsodium to implement a remote signing/verifying service that uses Ed25519ph. Client would hash the data to be signed with SHA512 and send the digest to a server, server would sign the result with Ed25519ph and return the signature. This way, if a client needs to sign a large blob (i.e. gigabytes), he doesn't need to upload the entire blob to the server.

Remote verify is less important but I would like to have it for symmetry, and to support cross-language clients for which I may not have good libsodium bindings.

Unfortunately, it doesn't look like libsodium exposes the internal `_crypto_sign_ed25519_detached()` and `_crypto_sign_ed25519_verify_detached()` functions in its public API. These are the functions that I would need to call if I am computing the sha512 by myself. Would it be possible to expose these functions in some way?

One possible idea: add 2 new public functions to `crypto_sign_ed25519.h`:

```
// signs a pre-hashed message.
// mhash is a pointer to the sha512 digest of the message to be signed.
// Other parameters are same as for crypto_sign_ed25519ph_final_create()
int crypto_sign_ed25519ph_prehashed(const unsigned char *mhash, unsigned char *sig, unsigned long long *siglen_p, const unsigned char *sk);

// verifies a pre-hashed message.
// mhash is a pointer to the sha512 digest of the message to be verified.
// Other parameters are same as for crypto_sign_ed25519ph_final_verify()
int crypto_sign_ed25519ph_verify_prehashed(const unsigned char *mhash, const unsigned char *sig, const unsigned char *pk);
```

ivmaykov:
Add a public API to sign/verify messages pre-hashed by the caller with Ed25519ph. This makes it possible to use libsodium to efficiently implement a remote signing service, for example (caller can hash a large message and send the digest only to the service for signing).

This fixes #789.

### DIFF
--- a/src/libsodium/crypto_sign/ed25519/sign_ed25519.c
+++ b/src/libsodium/crypto_sign/ed25519/sign_ed25519.c
@@ -81,7 +81,7 @@ crypto_sign_ed25519ph_final_create(crypto_sign_ed25519ph_state *state,
 
     crypto_hash_sha512_final(&state->hs, ph);
 
-    return _crypto_sign_ed25519_detached(sig, siglen_p, ph, sizeof ph, sk, 1);
+    return crypto_sign_ed25519ph_prehashed(ph, sizeof ph, sig, siglen_p, sk);
 }
 
 int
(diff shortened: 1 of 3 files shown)


jedisct1:
A caller can already prehash the file content and get that hash signed. This allows you to choose the hash function, and use standard EdDSA instead of the pH variant.

I'm really not convinced that introducing a new API is required.

ivmaykov:
The advantage, in my mind, is that Ed25519ph is a standardized algorithm with an RFC backing it and multiple implementations (including hardware implementations in at least one popular HSMs), so it has good interoperability properties. Doing a manual `sign(hash(message))` works, but it means the verifier has to perform the extra `hash(message)` step as well. Contrast with Ed25519ph, which still lets me do hashing client-side to avoid uploading gigabytes of data to the signing server, and verifier just has to call `verify(message)` with ed25519ph as the algorithm.

An alternative would be to expose the internal `_crypto_sign_ed25519_detached` and `_crypto_sign_ed25519_verify_detached` functions in the public API, maybe gated behind an optional build-time configuration flag. What do you think?

jedisct1:
Hi Ilya,

Doing `sign(hash(message))` is no different from what Ed25519ph does. You can still compute `hash(message)` on the client.

I understand that the ability to skip the initial hash when using Ed25519ph is something that can be useful for your specific use case, but I don't think this is something a lot of people will use. Yet, adding this API means that it will have to be supported forever, as well as implemented and supported in all bindings.

Since this doesn't solve a problem that cannot be solved already, I'd rather pass on this.

ivmaykov:
Well I think calling the non-exported function is technically possible (I have a PoC with a hard-coded function offset that I dug up using `objdump`), but difficult enough to do in practice (in a way that is portable across compilers/architectures/build modes) that copying the ed25519 implementation out of libsodium is probably simpler and safer. I would hate to do that just because 2 functions that implement exactly what I need are not exported :(

ivmaykov:
One more thought,

> Doing `sign(hash(message))` is no different from what Ed25519ph does. You can still compute hash(message) on the client.

This is not strictly true, since `ed25519(sha512(message) != ed25519ph(message)` , by design. So, using `ed25519(hash(message)` is conceptually the same, but produces signatures that are not compatible with `ed25519ph(message)` :(

jedisct1:
Yes, but the main purpose of the library is to provide operations, not specific algorithms. 

The library has too much bloat already, so I'm sorry, but new APIs should really be introduced only if they are strictly required to do something that was not possible before.

PR #799, opened 2019-01-14
Add -pthread to pkgconfig's Libs.private if enabled

bluca:
Allows static builds to correctly inherit the pthread dependency when
used with pkg-config --static --libs libsodium


jedisct1:
Thank you!

Shall this be backported to `stable`?

bluca:
Sure no problem, it's an easy cherrypick: https://github.com/jedisct1/libsodium/pull/800

PR #800, opened 2019-01-14
Add -pthread to pkgconfig's Libs.private if enabled [stable]

bluca:
Allows static builds to correctly inherit the pthread dependency when
used with pkg-config --static --libs libsodium

(cherry picked from commit 3933a7402c6614a2af45d9c89381760eb94f91e7)


jedisct1:
Is `Libs.private:` (empty content) fine with any `pkg-config` version?

bluca:
Yes, as far as I'm aware I've never seen that being a problem

PR #804, opened 2019-02-02
Relax most __attribute__ ((nonnull)) to allow 0-length inputs to be NULL

ivmaykov:
Justifications:
- crypto_(auth|hash|generichash|onetimeauth|shorthash)*:
  it's legal to hash or HMAC a 0-length message
- crypto_box*: it's legal to encrypt a 0-length message
- crypto_sign*: it's legal to sign a 0-length message
- utils:
  comparing two 0-length byte arrays is legal
  memzero on a 0-length byte array is a no-op
  converting an empty hex string to binary results in an empty binary string
  converting an empty binary string to hex results in an empty hex string
  converting an empty b64 string to binary results in an empty binary string
  converting an empty binary string to b64 results in an empty b64 string
  sodium_add / sodium_sub on zero-length arrays is a no-op

For the functions declared in utils.h, I moved the logic into private functions that
have the __attribute__ ((nonnull)) check, but they are only called when the
corresponding length argument is non-0. I didn't do this for the hash/box/sign
functions since it would have been a lot more work and quite a large refactor.


jedisct1:
Thanks!

The private functions are probably not be required, especially since short-circuiting the actual functions can introduce bugs (ex: returning `hex` in `sodium_bin2hex()` is not correct, since we still should fill `hex[0]` with `\0`).

We'd better make sure that the implementations don't call libc functions for which a `NULL` pointer would be an UB instead.

Also, in 1.0.17, the test suite tries to avoid `NULL`, even for empty messages to respect the attributes. If these attributes are changed, we should update these tests to use `NULL` again.

ivmaykov:
One advantage of having the private functions is that calling the public function with `(NULL, len > 0)` will still trigger UBSAN, which might be a bit nicer to debug than a segfault. But it's not a very strong argument, so I would be fine with removing them. Let me know what you prefer.

Will update the PR with tests that pass in `(NULL, 0)` arguments in the next couple days.

ivmaykov:
@jedisct1 Removed private functions, updated / added some tests.
I haven't updated all the tests yet, but let me know if what I have so far is on the right track (in terms of style, coverage, etc).

jedisct1:
Thanks, Ilya!

Can you fix [these minor issues](https://travis-ci.org/jedisct1/libsodium/jobs/488305272#L3799) (around line 3799) as well?

ivmaykov:
Heh, didn't expect tis to be merged as-is, was going to add more tests and address the issues you pointed out. I'll open another PR for that when I get some free cycles.

PR #812, opened 2019-04-02
Ed25519: add EdDSA-Ed25519-Blake2b-512 signing variant

tini2p-project:
Copy of ref10 code, adding EdDSA-Ed25519-Blake2b-512 variant.

Replaces internal use of Sha-512 with Blake2b-512, protects against [Length Extension Attacks](https://en.wikipedia.org/wiki/Length_extension_attack).

### DIFF
--- a/src/libsodium/Makefile.am
+++ b/src/libsodium/Makefile.am
@@ -73,6 +73,9 @@ libsodium_la_SOURCES = \
 	crypto_sign/ed25519/ref10/open.c \
 	crypto_sign/ed25519/ref10/sign.c \
 	crypto_sign/ed25519/ref10/sign_ed25519_ref10.h \
+	crypto_sign/ed25519/blake2b/keypair.c \
+	crypto_sign/ed25519/blake2b/open.c \
+	crypto_sign/ed25519/blake2b/sign.c \
 	crypto_stream/chacha20/stream_chacha20.c \
 	crypto_stream/chacha20/stream_chacha20.h \
 	crypto_stream/chacha20/ref/chacha20_ref.h \
--- a/src/libsodium/crypto_sign/ed25519/blake2b/crypto_sign_ed25519_blake2b.h
+++ b/src/libsodium/crypto_sign/ed25519/blake2b/crypto_sign_ed25519_blake2b.h
@@ -1,0 +1,20 @@
+#ifndef sign_ed25519_blake2b_H
+#define sign_ed25519_blake2b_H
+
+void _crypto_sign_ed25519_blake2b_hinit(crypto_generichash_blake2b_state *hs,
+                                        const unsigned char *salt,
+                                        const unsigned char *personal);
+
+int _crypto_sign_ed25519_blake2b_detached(unsigned char *sig,
+                                          unsigned long long *siglen_p,
+                                          const unsigned char *m,
+                                          unsigned long long mlen,
+                                          const unsigned char *sk,
+                                          const unsigned char *personal);
+
+int _crypto_sign_ed25519_blake2b_verify_detached(const unsigned char *sig,
+                                                 const unsigned char *m,
+                                                 unsigned long long   mlen,
+                                                 const unsigned char *pk,
+                                                 const unsigned char *personal);
+#endif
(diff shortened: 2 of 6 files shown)


tini2p-project:
Wasn't sure of best way to add test vectors.

Could copy the ref10 tests, and replace signature with the ones produced by the Blake2b variant.

Let me know preferred way to go for tests, and I'll add them.

enkore:
> Replaces internal use of Sha-512 with Blake2b-512, protects against Length Extension Attacks.

The underlying hash function doesn't need to protect against length extension, because the hash itself is only used as the identity of the message, which isn't broken by length extension. You can of course easily generate a message which extends a signed message, but because you can't transform the extended hash into a valid signature it doesn't give you anything.

jedisct1:
Thanks for jumping in, @enkore !

tini2p-project:
Understood, I need this as a potential signing variant, and thought I would send the patch upstream.

Thanks for the quick response, and providing a crypto library that's very nice to work with/extend :)

jedisct1:
You would end up with a variant that's non standard, not compatible with anything, and has no benefits over the standard construction. I'm not sure this is worth it.

If you really need a variant, you may want to implement Shnorr signatures over Ristretto instead.

tini2p-project:
> If you really need a variant, you may want to implement Shnorr signatures over Ristretto instead.

We are currently discussing new signing variants, and I considered proposing XEdDSA, which is over Curve25519. libHydrogen is attractive for it's simplicity + Gimli.

The current variants being discussed are:

- EdDSA-Ed25519-Blake2b-512
- RedDSA-Ed25519-Blake2b-512 (non-standard variant w/ key/signature-blinding)
- EdDSA-Ed25519-Sha-512-Ctx (standard from RFC 8032)

Now I will look deeper into Schnorr over Ristretto, and come up with a proposal. Thanks for your feedback.

jedisct1:
What problem are you trying to solve?

tini2p-project:
> What problem are you trying to solve?

In I2P, there are a number of core data structures that are encrypted and/or signed, containing information about network nodes.

Distributed bootstrapping nodes ([floodfills](https://geti2p.net/en/docs/how/network-database#floodfill)) verify the signatures, and relay the data structures for other nodes.

New signature algorithms are being discussed for blinded identities/signatures, which make encrypted short-term contact structures ([lease sets](https://geti2p.net/en/docs/how/network-database#encrypted2)) possible.

From a discussion on IRC (ed = [Ed25519](https://tools.ietf.org/html/rfc8032), red = [Red25519](https://geti2p.net/spec/proposals/146-red25519)):

```
13:52 <~zzz> 1) ed->ed blinding, as tor does it, str4d says is flawed because the keys are not on the prime order subgroup, although afaik he has not confronted them about it to get a response
13:53 <~zzz> 2) ed->red blinding is flawed because the keys are not identically distributed, as required according to zcash. We support it to encrypt exsiting ed dests.
13:53 <~zzz> 3) red->red does not have either flaw, and is preferred
```

I also brought up your suggestion of using Schnorr over Ristretto, and am doing research to develop a proposal.

jedisct1:
Good luck. But I have a bad feeling about this.

tini2p-project:
> Good luck.

Thanks :)

> But I have a bad feeling about this.

Yeah, a number of people involved have expressed similar sentiments about many sharp edges.

Trying to be cautious, and "do it right".

tarcieri:
FYI, there are a few existing Ristretto Schnorr schemes. I think @str4d can clue you in for I2P purposes

tini2p-project:
> FYI, there are a few existing Ristretto Schnorr schemes.

Awesome, thanks for the heads-up.

> I think @str4d can clue you in for I2P purposes

Definitely, hoping to touch bases soon.

IvanAnishchuk:
I just wonder, is this implemention compatible with the one implemented in https://github.com/nanocurrency/nano-node/tree/master/crypto/ ? (Project used to be called raiblocks)

IvanAnishchuk:
Maybe this would help https://github.com/nanocurrency/nano-node/wiki/Design-features

PR #825, opened 2019-04-27
Fix placement of alignment specifier

Fraser999:
I guess this was the original intent for the alignment specifier?


jedisct1:
Thanks!

PR #830, opened 2019-05-08
Add incremental mode for shorthash / siphash2,4.

graydon:
This does a simple incremental mode for the shorthash / siphash2,4 implementation. Modeled on the generichash routines. Itch-scratching on our end -- we want a hashtable that can hash things without having to allocate a buffer and serialize the object. Performs better (and more evenly) if we can use a fixed-size buffer and scan/flush it as we pass over the object.

### DIFF
--- a/src/libsodium/crypto_shorthash/crypto_shorthash.c
+++ b/src/libsodium/crypto_shorthash/crypto_shorthash.c
@@ -32,3 +32,25 @@ crypto_shorthash_keygen(unsigned char k[crypto_shorthash_KEYBYTES])
 {
     randombytes_buf(k, crypto_shorthash_KEYBYTES);
 }
+
+int
+crypto_shorthash_init(crypto_shorthash_state *state,
+                      const unsigned char *k)
+{
+    return crypto_shorthash_siphash24_init(state, k);
+}
+
+int
+crypto_shorthash_update(crypto_shorthash_state *state,
+                        const unsigned char *in,
+                        unsigned long long inlen)
+{
+    return crypto_shorthash_siphash24_update(state, in, inlen);
+}
+
+int
+crypto_shorthash_final(crypto_shorthash_state *state,
+                       unsigned char *out)
+{
+    return crypto_shorthash_siphash24_final(state, out);
+}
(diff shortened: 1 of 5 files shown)


jedisct1:
Hi,

And thanks for your PR!

If such a change is made, it has to be made for all the variants (128 bit output, and/or reduced round version), both in the high level and low-level interfaces.

But ss you know, SipHash is designed for short inputs. For other cases, people are encouraged to use BLAKE2 instead.

So, I'm not sure that a multi-part API is worth adding.

This is the first time such a request is made even though SipHash was implemented 6+ years ago, so even though there are valid use cases for it, these don't seem to be common enough to justify it. It's also a very simple function to reimplement if this is really needed.

It can be revisited later if more people show interest for that interface. But libsodium already has too many functions that can be considered bloat, as they will be sticking around forever with very few projects actually using them. So, I'm a little reluctant to add this.

graydon:
Totally understand & appreciate the motive to avoid useless functions.

If I can offer an attempt at a persuasive counterargument: SipHash was designed to be _cheap enough_ to use for small inputs (i.e. not rely on amortizing costs over long inputs), but there's nothing in that design that makes it work poorly on larger inputs. I.e. its design is not in any way _exclusive_ to small inputs.

Whereas it definitely _was_ designed for use as a key-hashing function in hashtables, and that's what we're using it for here. Typically one does not control the key size of a hashtable: it has small keys so needs to use SipHash, but it also has some number of larger keys (potentially lots -- user-controlled). To do what you're suggesting here would both make for weird code -- the hash function would have to do a size estimate and then pick one of two hashes -- it would also make the hashtable have a weird performance discontinuity at whatever threshold one picks to switch between SipHash and Blake2b.

I should also note that several implementations of SipHash are structured this way (including the one in the Rust stdlib):

https://github.com/vcatechnology/siphashsum
https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#3202
https://github.com/emboss/siphash-c
https://github.com/dchest/siphash
https://github.com/whitfin/siphash-cpp

(In fact I was using one of these for our streaming case on monday and my boss asked me to contribute a streaming variant fo sodium to avoid adding a non-sodium crypto dependency, so..)

stanford-scs:
The reason this change is important is that these kinds of hash computations often occur in inner-loop type situations, where you'd like to be able to hash basically a scatter-gather `iovec` like structure rather than allocate a buffer and flatten out whatever it is you are hashing.

graydon:
(ping @jedisct1 -- just realized github likely does not notify of comments on closed issues..)

PR #832, opened 2019-05-11
Fix libsodium version detection in CMake recipe

xmrdsc:
1. Fixes version detection, did not work for me (using CMake 3.14.2)
2. Exposes a new variable `sodium_VERSION`

Please note I don't actually know how CMake works, so double check if this PR makes sense.

### DIFF
--- a/contrib/Findsodium.cmake
+++ b/contrib/Findsodium.cmake
@@ -21,6 +21,7 @@
 #   sodium_INCLUDE_DIR
 #   sodium_LIBRARY_DEBUG
 #   sodium_LIBRARY_RELEASE
+#   sodium_VERSION
 #
 #
 # Furthermore an imported "sodium" target is created.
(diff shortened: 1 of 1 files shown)


jedisct1:
You know `cmake` way better than I do, but the presence of that `sodium_VERSION` definition doesn't seem to be a good thing.

`SODIUM_VERSION_STRING` is a free-form, informational string describing the installed package. "devel-1.0.19beta-packaged_for_redhat" could be a valid string.

This is only supposed to be used for debugging purposes. And even then, it only makes sense if the library is statically linked.

Feature detection at compile time should either check for available functions (isn't it what `cmake` was designed for?), or use `SODIUM_LIBRARY_VERSION_MAJOR`, `SODIUM_LIBRARY_VERSION_MINOR` and `SODIUM_LIBRARY_MINIMAL`.

`sodium_VERSION` should be removed and replaced with definitions for these.

xmrdsc:
> This is only supposed to be used for debugging purposes. 

Was exactly my use case. I like the extra output in the 'detection stage' (not sure what the right word is).

> `sodium_VERSION` should be removed and replaced with definitions for these.

Sounds good.

Thanks

PR #833, opened 2019-05-11
Override print and printErr

bas-d:
This prevents libsodium from breaking when `drop_console=true` is used with uglify. It still shows an `unhandled promise rejection` error on loading. This can either be fixed by overriding `Module.instantiateWasm` with a lot of copied code from emscripten, or making some adjustments to emscripten. See https://github.com/emscripten-core/emscripten/pull/8558 for the progress of an attempt to accomplish the latter :). If it doesn't work out we can always go with option one (or accept that it shows an error..).

Related to https://github.com/jedisct1/libsodium.js/issues/196 and https://github.com/jedisct1/libsodium.js/pull/198

### DIFF
--- a/dist-build/emscripten.sh
+++ b/dist-build/emscripten.sh
@@ -100,10 +100,14 @@ if [ "$DIST" = yes ]; then
       var Module = _Module;
       Module.onAbort = reject;
       Module.print = function(what) {
-        console.log(what);
+        if (console != null) {
+          console.log(what);
+        }
       }
       Module.printErr = function(what) {
-        console.warn(what);
+        if (console != null) {
+          console.warn(what);
+        }
       }
       Module.onRuntimeInitialized = function() {
         try {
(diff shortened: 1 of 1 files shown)


jedisct1:
This is great progress!

The `console` object may not always be available though, so maybe we want to check that it exists before calling its methods.

bas-d:
Ah yeah good point, I'll add that tomorrow. However, they do get dropped because of `drop_console` in current version of libodium.js. I've just put them in there so messages from emscripten are still shown when you disable uglify and `drop_console` in particular.

jedisct1:
That would throw an exception. You check that `typeof(console) !== 'undefined'`.

bas-d:
Are you sure? 
`console=undefined;if (typeof(console) !== 'undefined') { console.log("No error") };if (console != null) { console.log("Also no error") }`
doesn't throw in my browser. But happy to change it if you want!

jedisct1:
In your browser, `window.console` was previously defined. But in Javascript, you can't use a variable without having defined it first.

You can verify this by accessing `konsole` or some random non-declared name.

bas-d:
Cool, didn't know that, thanks. I'll update!

PR #839, opened 2019-06-08
Namespace private symbols to prevent static linking conflicts

orivej:
I wanted to statically link libsodium 1.0.18 and argon2 into an application and noticed that they are in conflict over such symbols as `argon2_ctx` and `initialize`. This PR fixes this and moves other public symbols into crypto_* namespaces. (randombytes and ed25519_ref10 are the exception. If you like I can move them too.)

Please review the commits individually. (My application builds after the first two.)

### DIFF
--- a/src/libsodium/crypto_generichash/blake2b/ref/blake2.h
+++ b/src/libsodium/crypto_generichash/blake2b/ref/blake2.h
@@ -23,19 +23,23 @@
 #include "crypto_generichash_blake2b.h"
 #include "export.h"
 
-#define blake2b_init_param crypto_generichash_blake2b__init_param
-#define blake2b_init crypto_generichash_blake2b__init
+#define blake2b_init_param crypto_generichash__blake2b_init_param
+#define blake2b_init crypto_generichash__blake2b_init
 #define blake2b_init_salt_personal \
-    crypto_generichash_blake2b__init_salt_personal
-#define blake2b_init_key crypto_generichash_blake2b__init_key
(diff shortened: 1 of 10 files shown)


jedisct1:
These are private symbols. Why do they need to be namespaced?

This PR also includes changes that have nothing to do with namespacing.

I'm not sure what problem this is fixing, but this seems to be a very specific problem that nobody else ever reported in the past 6 years. So I'd rather keep things simple rather than introduce additional constraints to maintain that code.

orivej:
When the library is built statically (`./configure --enable-static`) only `static` symbols are private. I had the impression that libsodium supports static linking because you distribute static libraries for Windows, and the purpose of b381f42d178b2568b635e616e7b3738fc1f74288 seems to allow using `libsodium` and `blake2b` in the same application. In that case the library should not define non-static symbols that are also defined in other libraries; in particular, it should rename argon2 symbols or make them static.

orivej:
I have noticed that you did some renaming in https://github.com/jedisct1/libsodium/compare/d855d308261a7059536044b9c61220244711ae38...7214dff083638604cd48e5c9ffc5704460192794 . It looks like your purpose was symbol hygiene within libsodium, and that you either do not want to support static linking or do not understand its constraints. If the latter is the case, I can provide an example that shows how libsodium and libargon2 can not be used together statically.

orivej:
Thank you for fixing this! Your symbol renamer is working perfectly.

PR #840, opened 2019-06-24
Update utils.c

mpex:
I noticed that the shielding_key is not used in sodium_mshield() (only filled in crypto_generichash())
Is the wrong key used in crypto_stream_xor?


jedisct1:
Thanks!

PR #853, opened 2019-08-06
Set NODEJS_CATCH_EXIT=0 in emscripten LDFLAGS

willclarktech:
Closes https://github.com/jedisct1/libsodium.js/issues/205

At some point it might also be nice to do the same for the recently supported `NODEJS_CATCH_REJECTION` flag: https://github.com/emscripten-core/emscripten/pull/9061/files#diff-43d6bd0c5e0f9bc940befa68ab214b4dR527-R532


webmaster128:
@jedisct1 would it be possible to cherry-pick this commit on the `stable` branch? It is important for us to allow the use of libsodium.js with nodejs 12, which is getting increasingly popular.

PR #866, opened 2019-09-11
Fix for new emscripten LLVM backend

buu700:
Got another libsodium.js fix here. As I [discovered last night](https://github.com/emscripten-core/emscripten/pull/9407#issuecomment-530156742), apparently asm.js with the new LLVM backend no longer initializes synchronously.

This shouldn't matter now unless you're building with emscripten `latest-upstream`, but it sounds like the new backend will become the default pretty soon.

### DIFF
--- a/dist-build/emscripten.sh
+++ b/dist-build/emscripten.sh
@@ -116,20 +116,28 @@ if [ "$DIST" = yes ]; then
         }
       };
       Module.useBackupModule = function() {
-        var Module = {};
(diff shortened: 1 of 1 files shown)


jedisct1:
Thank you for keeping track of everything happening with emscripten!

jedisct1:
Not merging to `-stable` until LLVM becomes the default backend.

buu700:
Not a problem! And cool, sounds good.

PR #874, opened 2019-09-24
add missing initialization in randombytes_random for emscripten

korb-monument:
randombytes_random tried to call Module.getRandomValue without ensuring that it was actually initialized, fixed by calling randombytes_stir which provides this initialization in the emscripten build.


jedisct1:
Why? `randombytes_stir()` is already called in `sodium_init()`.

PR #876, opened 2019-09-26
Add support for Particle targets

Dan-Kouba:
Just want to start a discussion here about the best way to have this support added to the repo.  I have successfully built a version of this library that compiles for Particle targets, but it requires a decent amount of finagling of the directory structure to match what our build system wants, as well as adding some defines to files to avoid custom compiler directives.  I've asked our team to look into adding more customization options to the build system, but until then I'm working with the system as it works today.

I look to the maintainers to suggest a good implementation for "building" the library (more just reformatting) within the existing system.

Otherwise my edits to the random number generator implementation are based on the [ESP32 port](https://github.com/espressif/esp-idf/tree/master/components/libsodium) of this repository.

Interested to hear any ideas!

### DIFF
--- a/dist-build/particle.sh
+++ b/dist-build/particle.sh
@@ -1,0 +1,54 @@
+#! /bin/sh
+
+# Build the Particle library structure
+[ -d "builds/particle" ] && rm -rf builds/particle
+mkdir builds/particle
+mkdir builds/particle/examples
+mkdir builds/particle/src
+
+# Copy over the libsodium code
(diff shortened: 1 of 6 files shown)


jedisct1:
Hi Dan,

And thanks for porting this library to Particle targets!

Unfortunately, this is not something I will be able to test (does the test suite even run?). So, this is something that would be better maintained separately, by people who actually have the hardware, are familiar with it, and can do a way better job than me at making sure that everything always works as expected.

PR #884, opened 2019-10-11
Optimisation to succeed fast when checking signature scalar is reduced.

isislovecruft:
Hi! Thank you for maintaining libsodium!

This provides a minor optimisation for ed25519 signature verification, when used without the -DED25519_COMPAT feature, to strictly check for a fully reduced scalar, `s`, component in variable time by first checking that the most significant *four* bits are unset, and only if any of them are set proceed to the `sc25519_is_canonical` check which performs the full reduction.  This should result in succeeding fast for the check on roughly half of all well-formed, canonicalised signatures.

This is safely backwards compatible with the previous implementation of strict checking for signature scalars.

This is effectively equivalent to https://github.com/dalek-cryptography/ed25519-dalek/pull/99

### DIFF
--- a/src/libsodium/crypto_sign/ed25519/ref10/open.c
+++ b/src/libsodium/crypto_sign/ed25519/ref10/open.c
@@ -28,8 +28,11 @@ _crypto_sign_ed25519_verify_detached(const unsigned char *sig,
         return -1;
     }
 #else
-    if (sc25519_is_canonical(sig + 32) == 0 ||
-        ge25519_has_small_order(sig) != 0) {
+    if (sig[63] & 240 &&
+        sc25519_is_canonical(sig + 32) == 0) {
+      return -1;
+    }
+    if (ge25519_has_small_order(sig) != 0) {
         return -1;
     }
     if (ge25519_is_canonical(pk) == 0 ||


isislovecruft:
Uh oh spaghettios. I thought I had broken your CI, but on closer inspection it looks like something else was already failing in `master`. Please let me know if you need a rebase or any modifications!

kchalkias:
@isislovecruft thanks for the optimization trick. Note that this will succeed fast on
`2^252 / (2^252 + 27742317777372353535851937790883648493)` of all well-formed, canonicalized signatures, thus, not only on half of them but almost always!

jedisct1:
Thank you!

real-or-random:
@kchalkias I haven't verified what you said but if you're right then you could have a variant of the scheme that always rejects if the one of the four bits is set, and just accept the fact that correctly produced signatures verify only with overwhelming probability.

kchalkias:
@real-or-random, I think the rejecting policy is not equivalent and definitely not aligned to the RFC. It seems that you cannot reject a signature if any of the four bits is set, because some signatures (roughly 2^124) with the 4th bit set are valid and shouldn’t be rejected.

Also, especially in blockchains, a succeed fast approach is preferred, as normally miners or validators have incentives to add correct blocks and this signature will be verified by many.

real-or-random:
> @real-or-random, I think the rejecting policy is not equivalent and definitely not aligned to the RFC. It seems that you cannot reject a signature if any of the four bits is set, because some signatures (roughly 2^124) with the 4th bit set are valid and shouldn’t be rejected.

(Assuming your probabilities are correct), my point is that honest signers will produce these signatures only with negligible probability. Going a step further, it should actually be computationally hard for a malicious signer to find a valid signature with a scalar with one of the four bits set.

kchalkias:
@real-or-random actually it’s very easy for a malicious signer to generate a non canonical S value.

isislovecruft:
@real-or-random The gist of the problem with the fourth bit being set/unset is that the order of the basepoint is `2^252 + 27742317777372353535851937790883648493 = 2^252 + 2^124 + 6474669844813699569391024826398135277`, which is basically like having a 2^252.5 bit integer where there are `2^124 + 6474669844813699569391024826398135277` possible valid canonical scalars which have the fourth bit set. It's the annoying extra half-bit in there that makes the full reduction check necessary if the fourth bit is set. Most ed25519 implementations—in order to maintain compatibility with the `ref10` legacy reference implementation—is to only check the first three bits via `if (sig[31] & 224) { return -1; }`, however this has two problems: 1) it's not compatible with the (written a decade later) RFC, and 2) all the `2^253 - 2^252 - 6474669844813699569391024826398135277` _invalid_ scalars which have the fourth bit set.

real-or-random:
Thanks, this makes sense.

jedisct1:
As a side note, maintaining compatibility with legacy reference implementations doesn't seem worth it.

Strict checking of the scalar was implemented in libsodium circa 2014, way before the RFC. 

Number of reports about applications having been broken due to that change so far: zero.

`ED25519_COMPAT` is undocumented, and was added just in case some application would depend on the legacy behavior. But I'm not aware of anything that ever used that compatibility flag. And now is probably a good time to retire it.

PR #887, opened 2019-10-23
Simplify WASI build using wasienv

syrusakbary:
This PR simplifies the build system using [wasienv](https://github.com/wasienv/wasienv).

This can be incredibly useful to allow contributors to generate the WASI WebAssembly files in their own system easily (since `wasienv` is already available in macOS, Linux and soon Windows!)

Fun fact: [the WASI test wrapper](https://github.com/jedisct1/libsodium/blob/master/test/default/Makefile.am#L507-L509) could also be safely removed (I didn't do it in this PR just in case it's useful to test with more runtimes, but we can add this feature in wasienv if you think it can be useful!)

### DIFF
--- a/dist-build/wasm32-wasi.sh
+++ b/dist-build/wasm32-wasi.sh
@@ -10,8 +10,15 @@ export PREFIX="$(pwd)/libsodium-wasm32-wasi"
 mkdir -p $PREFIX || exit 1
 
 export CFLAGS="-DED25519_NONDETERMINISTIC=1 -O2"
+export LDFLAGS="-s -Wl,--no-threads"
 
 make distclean > /dev/null
+
+# This is necessary for Frankeinstein Linux systems, where people mix a base
+# system containing old tools (here: old autotools) with backported or
+# hand-compiled recent tools (here: LLVM)
+grep -q -F -- 'wasi' build-aux/config.sub || \
+  sed -i -e 's/-nacl\*)/-nacl*|-wasi)/' build-aux/config.sub
 
 if [ "x$1" = "x--bench" ]; then
   export BENCHMARKS=1
--- a/dist-build/wasm32-wasi.sh
+++ b/dist-build/wasm32-wasi.sh
@@ -1,36 +1,17 @@
 #! /bin/sh
 
-if [ -z "$WASI_LIBC" ]; then
-  for path in /opt/wasi-libc /opt/wasi-sysroot; do
-    if [ -d "$path" ]; then
-      export WASI_LIBC="$path"
-      break
-    fi
-  done
-fi
-if [ -z "$WASI_LIBC" ]; then
-  echo "Set WASI_LIBC to the path to the WASI libc sysroot" >&2
+if ! command -v wasienv >/dev/null; then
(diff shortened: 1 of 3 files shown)


jedisct1:
`wasienv` is nice, but the LLVM versions available in Homebrew and in most operating systems can compile to WebAssembly without having to side-install another instance of the LLVM toolchain.

This is the setup I use (which has the benefit of also working on BSD and ARM) and that, according to the feedback we had on Lucet, most people seem to use.

Switching to `wasienv` for Azure Pipelines makes sense, though, as the currently available Linux distributions are pretty old (may this is not the case any more? I haven't checked recently).

But for regular builds, with the `wasm32-wasi.sh` script, I'd rather stick to using the system LLVM install.

dist-build/wasm32-wasi.sh
```diff
@@ -1,37 +1,18 @@
...
-export LDFLAGS="-s -Wl,--no-threads"
-export NM="llvm-nm"
-export AR="llvm-ar"
-export RANLIB="llvm-ranlib"
-export STRIP="llvm-strip"
+export CFLAGS="-DED25519_NONDETERMINISTIC=1 -O2"
 
 make distclean > /dev/null
 
-grep -q -F -- 'wasi' build-aux/config.sub || \
```
jedisct1: This is necessary for Frankeinstein Linux systems, where people mix a base system containing old tools (here: old autotools) with backported or hand-compiled recent tools (here: LLVM). And unfortunately, this is a very common setup.

dist-build/wasm32-wasi.sh
```diff
@@ -1,37 +1,18 @@
...
 
-export PATH="/usr/local/opt/llvm/bin:$PATH"
-
 export PREFIX="$(pwd)/libsodium-wasm32-wasi"
 
 mkdir -p $PREFIX || exit 1
 
-export CC="clang"
-export CFLAGS="-DED25519_NONDETERMINISTIC=1 --target=wasm32-wasi --sysroot=${WASI_LIBC} -O2"
-export LDFLAGS="-s -Wl,--no-threads"
```
jedisct1: `--no-threads` works around a bug in LLVM 8. Ubuntu still ships with LLVM 8 and that bug, so we still need it.

syrusakbary:
@jedisct1 feedback addressed!

syrusakbary:
> Switching to wasienv for Azure Pipelines makes sense, though, as the currently available Linux distributions are pretty old (may this is not the case any more? I haven't checked recently).

Oh, perhaps I understood it incorrectly.
Do you want me to create another build script for azure pipelines?

I also think it can be possible for wasienv to reuse the local llvm if that's useful for this project. Would you like me to work on that?

jedisct1:
For CI, we can definitely install `wasienv` as a first step.

But we shouldn't require `wasienv` at all if we already have the required tools to compile to WebAssembly.

`wasienv`, if already installed, can be used to find the correct path to the LLVM toolchain and to `wasi-libc`.

If no usable compilation toolchain is found, and we are on one of the platforms supported by `wasienv`, we can suggest installing it.

I'll take care of this.

PR #894, opened 2019-11-11
new crypto_musig (Multi-Signature) algorithm

ndilieto:
Hello, I've taken the liberty of implementing the MuSig algorithm described in "Simple Schnorr Multi-Signatures with Applications to Bitcoin" by Gregory Maxwell, Andrew Poelstra, Yannick Seurin, and Pieter Wuille, https://eprint.iacr.org/2018/068.pdf

I've called it crypto_musig and I'd appreciate if you could review it. Is there any hope of having it included in libsodium?


jedisct1:
This is great, thank you!

As any new operation introduced into the library, it would be better to first maintain this as a distinct project.

This will provide an opportunity to see if projects are actually using it, how they use it, and how the API may be improved according to their feedback. Maybe other instantiations of musig are going to see more traction. This also provides an opportunity to write the documentation according to the same criteria.

For eventual integration into libsodium:

- Please do not change how existing operations work. `ristretto` APIs are not build in minimal mode because these are not a high-level APIs.
- Please respect the naming scheme used absolutely everywhere else in the library, which is also documented. Your specific instantiation of `musig` should probably have a `crypto_multisig_musigristretto255xblake2b` prefix. If a high-level API is ever defined, it will have  eventually a`crypto_multisig` prefix. But there are other multi-signature schemes, and even musig can be instianciated on different groups and with different hash functions. Setting a high-level API right away would be a bit premature.
- Please use the same style as the existing code.
- Can the current musig API be reused with other multi-signature schemes? All primitives within a given operation must share the same API.

So, this can't be merged now. But this implementation can definitely be useful. So, make it a distinct project and see if people adopt it!

ndilieto:
> This is great, thank you!
> 
> As any new operation introduced into the library, it would be better to first maintain this as a distinct project.

Ok thanks, will do.

>     * Please do not change how existing operations work. `ristretto` APIs are not build in minimal mode because these are not a high-level APIs.

Maybe I misunderstood, I thought MINIMAL excludes anything that is not needed by high-level functions. My thinking was crypto_musig needs ristretto so it must come out of MINIMAL. In fact, when ristretto isn't in MINIMAL, make check fails on musig when --enable-minimal is on because it can't link ristretto...
 
> 
>     * Please respect the naming scheme used absolutely everywhere else in the library, which is also documented. Your specific instantiation of `musig` should probably have a `crypto_multisig_musigristretto255xblake2b` prefix.

Can you please tell me where this is documented? I think I really need to read this.

> 
>     * Please use the same style as the existing code.
I've actually tried to do this but perhaps I failed, can you please tell me in which respect I didn't meet the existing style?

> 
>     * Can the current musig API be reused with other multi-signature schemes? All primitives within a given operation must share the same API.
> 

musig can use any hash and any prime order group, however I think it's going to be a tall order to design a unified api that works with other multi-signature schemes. From what I've read so far on the topic, every multisignature is rather unique, with its own special requirements. For example, musig is as far as I know the only proven secure scheme able to aggregate the public key. On the other hand, it needs a three-stage protocol including two steps for sharing the nonces securely among participants. 

Thanks for your quick feedback. Would you have any comments on the musig API I coded? There is an example in test/default/musig.c

patrickkh7788:
I will try use this new projects.  and I hope it can merge into libsodium ASAP.

@ndilieto is this also support verifiable random function ?

jedisct1:
Hi @ndilieto ,

`crypto_multisig_musig*` wouldn’t be a high-level API, so just like ristretto, this is something that would require a full build, not a minimal one.

Some information on the internal code conventions are documented here: https://libsodium.gitbook.io/doc/internals

And for indentation, and general code style, we mostly use OpenBSD’s kernel normal form. Single-line comments are not used, variables are indented and sorted by type size, yoda-style comparisons is avoided. These are details, but having a consistent style makes it easier to review and maintain the code.

Having a cofactor > 1 shouldn’t be an issue for musig. Existing implementations use secp256k1 (a natural choice given what musig was mainly designed for) and edwards25519, with SHA256, SHA512 or BLAKE2B as a hash function. Ristretto isn’t a bad choice, but there doesn’t seem to be any other implementations having made it so far.

Being a generic composition is great, but right now, we don’t have any pairs of interoperable implementations. There are no standards, either formal or de-facto, which is a little bit annoying.
If musig-ristretto255-blake2b is implemented, but something else gains more traction, or use cases beyond bitcoin start discussions on multi-signatures on the CFRG list, that specific implementation (or musig itself) maybe become useless, rotten code that still requires to be maintained forever.
And in spite of being slower, BLS is actually likely to gain more traction, especially after the current work at standardizing pairing-friendly curves.

For all these reasons, and the fact that multi-signatures need more practical use cases than blockchains, I’d rather not merge an implementation in libsodium before a dedicated library has matured and appears to be something unlikely to need to be reconsidered shortly after.

PR #922, opened 2020-01-28
Added Catalyst Support

kappes:
Added Catalyst target to dist-build/iOS.sh, removed sim64 target b/c obsolete.

### DIFF
--- a/dist-build/ios.sh
+++ b/dist-build/ios.sh
@@ -55,6 +55,23 @@ PROCESSORS=${NPROCESSORS:-3}
 
 make -j${PROCESSORS} install || exit 1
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Did you test this?

Doesn't that break the simulator?

jedisct1:
Can you move these changes back to where the `x86_64 simulator` building part was, so that the changes are more readable?

kappes:
I'm working on a swift lib. Xcode 11 test target seem to need Catalyst support and that's working now as well as the sim. Seems like the Catalyst build replaces the x86_64.

jedisct1:
Thanks!

PR #942, opened 2020-03-30
prioritize less costly atomic ops over plain mutexes.

devnexen:
using active spinning over passive via pause asm.

### DIFF
--- a/src/libsodium/sodium/core.c
+++ b/src/libsodium/sodium/core.c
@@ -106,7 +106,7 @@ sodium_crit_leave(void)
     return 0;
 }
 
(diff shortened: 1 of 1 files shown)


jedisct1:
There is only one critical section, to protect `sodium_init()` from being calling multiple times. Speed is not really important here.

Still, [mutexes are faster than spinlocks](https://matklad.github.io/2020/01/04/mutexes-are-faster-than-spinlocks.html). And this is probably even more the case here considering the section it is protecting.

So I wouldn't change that. The spinlock is really a last resort solution and `nanosleep()` has the advantage of being available in WebAssembly/WASI.

Adding `yield` on `aarch64` is a good idea.

PR #946, opened 2020-04-17
Fix crypto_stream_chacha20_ietf tests

emilbayes:
Reviewing tests for `crypto_chacha20_ietf` I found that actually `crypto_chacha20` was being tested. I've updated the `.exp` file with the output, assuming it's correct (it seems so as this impl can successfully decrypt from other implementations)


jedisct1:
Good catch, thank you!

PR #979, opened 2020-07-03
added iOS xcframework build script

0x4cdc:
XCFramework will contain: 

- iOS slices
- iOS simulator both: i386 & x86_64 slices
- mac Catalyst


jedisct1:
Thank you!  ❤️

PR #1006, opened 2020-10-22
Define HAVE_LINUX_COMPATIBLE_GETRANDOM for Genode

ehmry:
Genode does not expose a /dev/random by default, which would trigger
a silent crash in sodium_init. Use getrandom() which is guaranteed
to be present.

### DIFF
--- a/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c
+++ b/src/libsodium/randombytes/sysrandom/randombytes_sysrandom.c
@@ -35,6 +35,8 @@
 # if defined(__FreeBSD_version) && __FreeBSD_version >= 1200000
 #  define HAVE_LINUX_COMPATIBLE_GETRANDOM
 # endif
+#elif defined(__GENODE__)
+#  define HAVE_LINUX_COMPATIBLE_GETRANDOM
 #endif
 #if !defined(NO_BLOCKING_RANDOM_POLL) && defined(__linux__)
 # define BLOCK_ON_DEV_RANDOM


ehmry:
[Genode](http://genode.org) is a post-UNIX OS with a compatibility libc derived from FreeBSD.

jedisct1:
Thanks!

Did you test this?

Genode seems to have a `get_random()` function, that has not only a different name than the Linux, but accepts 2 parameters, not 3.

`getrandom()` is also used by the internal RNG, so that one needs patching, too.

ehmry:
Sorry, I meant `getrandom`, not `get_random`. Its the same as the FreeBSD `getrandom`. Genode has no explicitly defined interface for randomness, but a grep of the sources for `get_random` may turn up some internal procedures in drivers and such.

jedisct1:
Thanks!

Does the entire test suite pass, even the ones using the internal RNG that wasn't patched?

Unless you can directly send a PR to get it tested on CI, would you mind helping me compile this for Genode and run the test suite?

I gave it a try using their toolchain, but couldn't get the correct command line to do it. The compilers have a prefix that doesn't match the standard triple, and even by manually trying to override the tool names, I didn't go very far :(

ehmry:
Yea, the official toolchain isn't compatible with anything beyond the internal build system.

I'm cross-compling so I haven't run the test suite, only tested an application compiled against libsodium.

If there is a Make target to compile the test suite without running it, I could the test the binaries in a VM and try to work that into CI.

[Build log](https://gist.github.com/ehmry/13828f1ab1514caec4fde221a22eef1d)

jedisct1:
I'm really awfully sorry, but this is a target that is very unlikely to receive much testing, and looks a little bit too complicated to add to CI.

It would probably be better to keep Genode-specific changes separately.

ehmry:
*cough* security theatre *cough*

PR #1009, opened 2020-11-04
Pack the NuGet package with .NET 5.0

ektrah:
A small update to use the .NET 5.0 SDK instead of the .NET Core 2.1 SDK for packing the NuGet package.

(Hope I got it right, since the GitHub Action only triggers for the *stable* branch...)


jedisct1:
Thank you!

PR #1024, opened 2021-01-27
Make sure config.sub and config.guess are executable

jonglezb:
This fixes Spack build on ppc64 and aarch64, see https://github.com/spack/spack/pull/21256

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -106,13 +106,13 @@ fi
 
   curl -sSL --fail -o config.guess \
     'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD' &&
-    mv -f config.guess build-aux/config.guess &&
+    chmod +x config.guess &&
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi,

The fact that they need to be executable is a little bit surprising.

The `configure` script doesn't call them directly, only via a shell command:

```sh
$SHELL "$ac_aux_dir/config.guess"
```

```sh
$SHELL "$ac_aux_dir/config.sub"
```

If the `SHELL` environment variable empty somehow? That would be the actual issue.

jonglezb:
Yes, I expect it would work for a normal build.

However, Spack has an autotools workaround where it tries to manually execute these scripts to check if they work (I have no idea what was the original issue that required this workaround). It's applied to all autotools-based packages, not just libsodium.

By the way, the upstream scripts are executable: https://git.savannah.gnu.org/gitweb/?p=config.git;a=tree

PR #1043, opened 2021-03-24
Fixed subset checking bug

wyatt-howe:
It seems I am the first to use the `generate-emscripten-symbols.sh` script to select a subset of the symbols smaller than the entire library.  The script does subset checking in the wrong direction, incorrectly asserting that the entire library is a subset. 
 It should be the opposite, as committed.  I also updated the big list of symbols and disabled a few functions that are no longer commonly used.


jedisct1:
`generate-emscripten-symbols.sh` generates the list of symbols to export for the current build. It's not designed to select a subset, except for the minimal mode.

Please do not change this, especially since you seem to be producing the symbols out of an antiquated version of the library.

PR #1045, opened 2021-04-02
make fe25519_sub inline

ReewassSquared:
why wasn't it inline?


ReewassSquared:
if there is a reason to not inline this function there should be an explanation within the block comment.

ReewassSquared:
I'm currently auditing x25519, ed25519 and curve25519 features in libsodium as part of another project. Additional PRs will be submitted for any bugs or inconsistencies found.

jedisct1:
Does it make any significant difference in speed on the supported platforms?

I'm not sure it does, but if that specific function had to be explicitly inlined, it should simultaneously be done in the 51-bit limbs counterpart.

Feel free to suggest inlining that function if actual benchmarks prove that to be necessary, and if the code size difference is not significant.

> if there is a reason to not inline this function there should be an explanation within the block comment.

Whoever maintains the code thought it was not necessary.

Issue #1050, opened 2021-04-12
[Question] How to properly clean up the critical section created by sodium_init()?


TSchick90:
Hey guys!

As the title says. I have some code checks in place which tell me about the critical section which is created during the `sodium_init()` but never really destroyed? Is there some `sodium_cleanup()` function or something I'm missing?

I would expect something like this in the core.c
```c
void sodium_crit_destroy()
{
   // if initialized, check _sodium_lock_initialized
   DestroyCriticalSection(&_sodium_lock);
} 
```

Stay save and best, Thomas

jedisct1:
There's nothing to do. sodium_init() doesn't do any heap allocations.

jedisct1:
Ah, I see what you are referring to. On windows, an `InitializeCriticalSection()` section is called by `sodium_init()`.
There's no `sodium_deinit()` function, so that value will not be deallocated. Not a big deal since it's only initialized once.

jedisct1:
If that worries you, the following diff should destroy the critical section on release.

```diff
--- a/src/libsodium/sodium/core.c
+++ b/src/libsodium/sodium/core.c
@@ -94,14 +94,23 @@ sodium_crit_enter(void)
 int
 sodium_crit_leave(void)
 {
+    LONG status = 0L;
+
     if (locked == 0) {
 # ifdef EPERM
         errno = EPERM;
 # endif
         return -1;
     }
+    while ((status = InterlockedCompareExchange(&_sodium_lock_initialized,
+                                                1L, 0L)) == 1L) {
+        Sleep(0);
+    }
     locked = 0;
     LeaveCriticalSection(&_sodium_lock);
+    DeleteCriticalSection(&_sodium_lock);
+    _sodium_lock = NULL;
+    InterlockedAnd(&_sodium_lock_initialized, 0L);
 
     return 0;
 }
```

Completely untested, and probably not something that's going to be merged.

TSchick90:
Ah yes, that I saw this on Windows would have been a useful information ... anyway, thanks for your answers :)

PR #1055, opened 2021-05-11
Add sodium_finalize()

cmb69:
`sodium_crit_enter()` lazily initializes a critical section on Windows
but that critical section is never deleted (`DeleteCriticalSection()`).
This is not an issue per se, since the critical section is released on
process termination, but Application Verifier complains about that
("Unloading DLL containing an active critical section.")

There is no way yet to delete this critical section, since it is held
in a file static variable, and there is no API to delete it.  Thus,
I suggest to introduce `sodium_finalize()` which can be called by
clients to delete the critical section on process termination.

The new API should be available on all supported platforms, although
I don't know whether similar issues exists elsewhere.  For now, it does
nothing on other platforms than Windows.

---

For PHP's sodium extension, it could be used like:
````.diff
 ext/sodium/libsodium.c | 1 +
 1 file changed, 1 insertion(+)

diff --git a/ext/sodium/libsodium.c b/ext/sodium/libsodium.c
index abde3b8994..2eefd569a1 100644
--- a/ext/sodium/libsodium.c
+++ b/ext/sodium/libsodium.c
@@ -655,6 +655,7 @@ PHP_MINIT_FUNCTION(sodium)
 PHP_MSHUTDOWN_FUNCTION(sodium)
 {
 	randombytes_close();
+	sodium_finalize();
 	return SUCCESS;
 }
 ````

### DIFF
--- a/src/libsodium/include/sodium/core.h
+++ b/src/libsodium/include/sodium/core.h
@@ -12,6 +12,9 @@ SODIUM_EXPORT
 int sodium_init(void)
             __attribute__ ((warn_unused_result));
 
+SODIUM_EXPORT
+void sodium_finalize(void);
+
 /* ---- */
 
 SODIUM_EXPORT
--- a/src/libsodium/sodium/core.c
+++ b/src/libsodium/sodium/core.c
@@ -106,6 +106,21 @@ sodium_crit_leave(void)
     return 0;
 }
 
+void sodium_finalize(void)
+{
+    LONG status = 0L;
+
+    while ((status = InterlockedCompareExchange(&_sodium_lock_initialized,
+                                                1L, 0L)) == 1L) {
+        Sleep(0);
+    }
+
+    if (status == 2) {
+        DeleteCriticalSection(&_sodium_lock);
+        InterlockedExchange(&_sodium_lock_initialized, 0L);
+    }
+}
+
 #elif defined(HAVE_PTHREAD) && !defined(__EMSCRIPTEN__)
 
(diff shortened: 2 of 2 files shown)


jedisct1:
For DLLs, the `DllMain()` function is apparently called with a DLL is unloaded: https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain

That may be another way to do it without requiring applications to call a new function.

cmb69:
> We can't really expect applications to call a deinit function at this point.

As it is now (with this PR), there would be no *need* to call a deint function. :)

[Deleting the critical section in sodium_crit_leave()](https://github.com/jedisct1/libsodium/issues/1050#issuecomment-817865691) appears to be inefficient (would require to initialize it over and over again).

I thought about DllMain() as well, but I'm not sure if there won't be issues, because there is a long aritcle about [what not to do in DllMain()](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices).

Anyhow, thanks for your swift reply! I'll have a closer look at doing the deletion in DllMain(), and update the PR respectively.

jedisct1:
It not being super efficient shouldn't be an issue. This is only used to protect against multiple parallel executions of `sodium_init()`.

cmb69:
> It not being super efficient shouldn't be an issue. This is only used to protect against multiple parallel executions of sodium_init().

Ah, right! Anyhow, just submitted PR #1058, which might be preferable.

PR #1058, opened 2021-05-14
Delete critical section on DLL_PROCESS_DETACH

cmb69:
To explicitly release all allocated resources on shutdown on Windows,
we delete the critical section on `DLL_PROCESS_DETACH`.  We do not
employ any locking to avoid any potential deadlock.  In case of normal
DLL unloading there is no need to, and in case of forced unloading all
bets are likely off anyway.


cmb69:
Thank you!

PR #1063, opened 2021-05-21
Define randombytes_set_implementation argument to be const

jwf:
Discourage developers from creating writable function pointers. If someone insists on having their implementation spec be writable, it's easy to cast to const, but if someone does the right thing and defines it as a const, it's harder to go the other way.

### DIFF
--- a/src/libsodium/include/sodium/randombytes.h
+++ b/src/libsodium/include/sodium/randombytes.h
@@ -53,7 +53,7 @@ SODIUM_EXPORT
 int randombytes_close(void);
 
 SODIUM_EXPORT
-int randombytes_set_implementation(randombytes_implementation *impl)
+int randombytes_set_implementation(const randombytes_implementation *impl)
             __attribute__ ((nonnull));
 
 SODIUM_EXPORT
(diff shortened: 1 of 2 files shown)


jedisct1:
If my memory serves me correctly, this was intentional, due to some compilers not supporting constant function pointers.

jedisct1:
There wouldn't be anything wrong with passing a function pointer and change it later. I can see applications doing this intentionally, hardly by accident.

So, I'd rather avoid a breaking change here

cemeyer:
> If my memory serves me correctly, this was intentional, due to some compilers not supporting constant function pointers.

If we need to support broken compilers, we could conditionally define a macro that evaluates to `const` on conventional compilers and to nothing on broken compilers.  Would that be an ok workaround for this concern?  (Which compilers do we support that have this deficiency?)

> There wouldn't be anything wrong with passing a function pointer and change it later. I can see applications doing this intentionally, hardly by accident.
> 
> So, I'd rather avoid a breaking change here

`const` in this position doesn't prevent _applications_ from defining and modifying non-const vtables; it just promises that _libsodium_ won't modify the vtable via the application-provided pointer.  I don't believe this is a breaking change for applications using non-const storage for this vtable.

jedisct1:
Definitely better with the new commit message!

PR #1067, opened 2021-05-25
Add linux-arm64 binary to the generated nuget package

enclave-alistair:
Uses cross-compilation to run make; make check uses qemu to execute the tests on virtualised ARM64.

Note that I couldn't get the qemu-run make check to execute when running inside a docker container, so the arm64 job runs directly on the runner.

I've tested this process works on stable (see last run at https://github.com/enclave-alistair/libsodium/actions/runs/874111971).

Once the existing Autoconf build failure on master is resolved, I will test on master.

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -48,6 +48,50 @@ jobs:
       with:
         name: build-linux-x64
         path: .libsodium-build/lib/libsodium.so
+
+        
+  build-linux-glibc-arm64:
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Set up build environment
+      run: |
+        export DEBIAN_FRONTEND=noninteractive
+
+        cat <<-EOF | sudo tee /etc/apt/sources.list.d/arm64.list >/dev/null
+        deb [arch=arm64] http://ports.ubuntu.com/ focal main restricted
+        deb [arch=arm64] http://ports.ubuntu.com/ focal-updates main restricted
+        deb [arch=arm64] http://ports.ubuntu.com/ focal universe
+        deb [arch=arm64] http://ports.ubuntu.com/ focal-updates universe
+        deb [arch=arm64] http://ports.ubuntu.com/ focal multiverse
+        deb [arch=arm64] http://ports.ubuntu.com/ focal-updates multiverse
+        deb [arch=arm64] http://ports.ubuntu.com/ focal-backports main restricted universe multiverse
+        EOF
+
+        sudo sed -i 's/deb h/deb [arch=amd64] h/g' /etc/apt/sources.list
+
+        sudo dpkg --add-architecture arm64
+
+        sudo apt-get update && sudo apt-get install -y build-essential qemu-user qemu-user-static gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libstdc++6:arm64
+
+    - uses: actions/checkout@v1
+    - name: configure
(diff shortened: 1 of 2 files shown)


jedisct1:
Thank you!

enclave-alistair:
@jedisct1, can I ask if you have an anticipated approx date for a 1.0.19 push of the NuGet package onto nuget.org that might include the new binary?  The 1.0.18 version is from 2020, and I'd prefer to use a released package if possible.

ektrah:
It's also possible to push a 1.0.18.1. I can help with that; @jedisct1 just needs to hit the button at the end.

jedisct1:
How do I push the button?

I must confess that I didn't follow what had to be done now that we don't use the Docker image any more.

I thought the NuGet things were automatically following the `stable` branch. But apparently not?

What should I do?

enclave-alistair:
Without being too forward @jedisct1 / @ektrah, if you update the package version in https://github.com/jedisct1/libsodium/blob/stable/packaging/dotnet-core/libsodium.pkgproj#L12, GitHub Actions will build you a new `.nupkg` file with the right version and contents, available in the build artifacts.

The manual part should just involve uploading the released package to nuget.org (I couldn't see an automated step for that in your workflow).

ektrah:
@enclave-alistair Would it be possible to also include linux-arm32 with low effort?

And it would be really great if we could test the `nupkg` on both linux-arm64 and linux-arm32 [as we do for linux-x64](https://github.com/jedisct1/libsodium/blob/d1a5aaecf06fe387ef9acb81967282d0628d804c/.github/workflows/dotnet-core.yml#L186-L210).

enclave-alistair:
@ektrah, might be worth wildcarding out the version number in the `dotnet-core.yml`, just to save changing it in one place.

I can take a look at the `linux-arm` support (.net have used `-arm` for the 32-bit suffix), which would really be an armhf make target for libsodium itself, yes?

I'm not certain in my head on how we build and run the dotnet test project on ARM; with any luck the QEMU emulator might be able to run the dotnet SDK on an ARM architecture, but a bit unsure exactly.

enclave-alistair:
OK, raised another PR with the armhf changes, pretty straightforward.

If it's OK, I'll take a look at executing the nupkg tests on those platforms separately, because I suspect there'll some annoyances there.

ektrah:
> @ektrah, might be worth wildcarding out the version number in the `dotnet-core.yml`, just to save changing it in one place.

That would be nice to have. Last time I checked, GitHub Actions didn't support wildcards in that position.

> I can take a look at the `linux-arm` support (.net have used `-arm` for the 32-bit suffix), which would really be an armhf make target for libsodium itself, yes?

Awesome, thanks! I guess .NET on linux-arm means a Raspberry PI in practice? Is that armhf?

> If it's OK, I'll take a look at executing the nupkg tests on those platforms separately, because I suspect there'll be some annoyances there.

Sure. I think it would be good to test the `nupkg` in *some* way before uploading it to nuget.org (e.g. just once manually). Unfortunately, the only ARM device I have is a Cortex-M0, so I'm afraid I'm not much help here.

enclave-alistair:
Since there'll only be one .nupkg file in the Release folder, you can just specify `.libsodium-pack/bin/Release/*.nupkg`, and it'll work I believe.

Yeah, `linux-arm` is mostly just Raspberry Pi ARMv7 and some IoT devices.

I'll spin up an AWS Graviton instance tomorrow for the ARM64 test, that should be pretty straightforward. A bit trickier for the 32-bit, but I may have a spare raspberry pi available, I'll figure something out.

PR #1068, opened 2021-05-25
Define randombytes_set_implementation argument to be const.

jwf:
Allow applications to use const vtables. The promise here is that libsodium will not modify the application's vtable. Of course, libsodium does not and will not modify the application provided vtable, so this is just being explicit about an existing property. Applications providing non-const vtable pointers will have those pointers implicitly coerced to const in the function invocation, without any code changes in the application.

(Without this change, applications with const vtables must use kludges like casting-through-uintptr_t to strip the const qualifier from the vtable pointer.)


jedisct1:
Thank you!

PR #1073, opened 2021-06-13
core.c: Make _sodium_crit_init static

Alcaro:
Found by compiling under MinGW with -Wmissing-declarations.

Is Windows XP still relevant to libsodium? If no, I'll be happy to replace that CRITICAL_SECTION with an SRWLOCK <https://docs.microsoft.com/en-us/windows/win32/sync/slim-reader-writer--srw--locks>, which will allow deleting both this function and DllMain.


jedisct1:
Thank you!

PR #1099, opened 2021-08-21
Build x64 binaries in Ubuntu 16.04 to provide support for CentOS 7 and Debian 9

### LINKED ISSUE #1075: libsodium 1.0.18.1 NuGet package is broken on some older platforms
***Note:** This issue is about the [libsodium NuGet package](https://www.nuget.org/packages/libsodium/1.0.18.1), not about libsodium itself.*

My libsodium binding uses the libsodium NuGet package and fails its test cases on some platforms since NuGet package version 1.0.18.1:

screenshot

In each case, the dynamic linker is not able to find or load libsodium.{dylib/so} from the nuget package. My guess would be that this is because libsodium.{dylib/so} has been compiled on a platform that is newer than the failing one.

@jedisct1 Feel free to assign this issue to me. I will investigate when I find some time. Many thanks for your support of the whole nuget stuff so far! 😃

cc @enclave-alistair

enclave-alistair:
Also added tests that run the linux-x64 tests inside a centos container to verify the builds work on those OS versions.

One caveat, this fix currently won't allow Debian 9 on ARM chipsets via the nuget package, which **is** a supported .NET 5 target.  (CentOS on ARM is not supported in .NET).

The reason for this is that I haven't been able to get the QEMU compilation for arm to run successfully inside a docker container (the tests don't pass). It might be possible somehow, but figured it might be best to get the CentOS x64 support in and go back for Debian on ARM?

It's worth noting that .NET 6 will [remove support for Debian 9](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md#linux) (and will require Debian 10+). So this may not be as big of a deal.

Fixes #1075


jedisct1:
Any idea on what tests are failing on ARM? Any hints in the `test/default/*.log` files?

enclave-alistair:
@jedisct1; all tests fail with `cannot execute binary file: Exec format error`.

I'm guessing that make check is building an arm binary, but then trying to run it directly rather than via QEMU. 

When I call configure in the build I run `./configure --disable-debug --prefix=$PWD/.libsodium-build --host=aarch64-linux-gnu`. Any thoughts on why that `--host` argument may not be making it through to the make check?

ektrah:
I still think that it would be really good if we could add back support for Ubuntu 16.04, CentOS 7, and Debian 9.

It seems this pull request is mergeable if we forgo the tests on ARM (like we do in #1106). Would that be possible?

jedisct1:
Is it something we can maintain?

Ubuntu 16.04 reached end of life in April 2021, so it may soon not be available in GitHub Actions any more.

ektrah:
It seems that Ubuntu 16.04 reached end of standard support in April 2021 [but won't EOL before April 2026](https://wiki.ubuntu.com/Releases). So I'd assume that it won't go away anytime soon (it's currently even officially [supported by .NET 6](https://github.com/dotnet/core/blob/main/release-notes/6.0/supported-os.md)). If it eventually does, we can still go "back" to a later version, right?

enclave-alistair:
Just a note that I'm happy to continue supporting/working on this PR (16.04 is still something I'd like to support), if someone is able to assist me getting `make check` to run on QEMU if I pass a `--host` argument. It may just be a makefile change, but to be honest I don't have the make experience to make such a change without breaking everything else...

ektrah:
May I propose that we skip `make test` on QEMU for now (which should bring this PR into a mergeable state, right?) and go ahead with the other bits?

jedisct1:
Yep, that sounds good.

enclave-alistair:
Sorry for the delay @ektrah and @jedisct1, just got back to this; pushed the change to disable `make check` on linux-arm64. 

Also had to change the windows build to use vs2022, since it looks like that's what's available on the `windows-latest` runner. Seems like the nuget build hasn't worked on `stable` for a little while.

See the action run on my fork: https://github.com/enclave-alistair/libsodium/runs/5213899558

enclave-alistair:
@jedisct1, I believe this is ready to merge and push a new nuget release?

jedisct1:
@enclave-alistair Great!

Changes usually go from `master` to `stable`, never the other way round.

There are a couple conflicts here, I'm going to look at these.

enclave-alistair:
Great; let me know if you want me to retarget master my side.

PR #1105, opened 2021-09-15
Portability improvements

jorgenpt:
Improve portability to aid in porting libsodium to modern game consoles:
- Don't assume that all non-wasm targets have `raise`, `SIGKILL`, etc
- Don't assume that all platforms that have `_SC_PAGESIZE` also have `sysconf` (I've filed a bug with the vendor about the lack of a `sysconf`, but it might not be forthcoming)
- Don't generate warnings if we don't have a way to dynamically determine page size of `PAGE_SIZE` is defined

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -594,7 +594,7 @@ AC_SUBST(CFLAGS_AESNI)
 AC_SUBST(CFLAGS_PCLMUL)
 AC_SUBST(CFLAGS_RDRAND)
 
-AC_CHECK_HEADERS([sys/mman.h sys/random.h intrin.h sys/auxv.h])
+AC_CHECK_HEADERS([sys/mman.h sys/param.h sys/random.h intrin.h sys/auxv.h])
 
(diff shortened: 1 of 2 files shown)


angt:
goodjob

jedisct1:
Thank you!

The check for `__wasm__` instead of `HAVE_RAISE` was intentional. With your changes, compilation to WASI fails:

```text
In file included from sodium/utils.c:13:
/opt/zig/lib/zig/libc/include/wasm-wasi-musl/signal.h:2:2:
#error "wasm lacks signal support; to enable minimal signal emulation
```

We need to check both :)

jorgenpt:
@jedisct1, ah! I assumed the header didn't exist, thank you for addressing that. Is the `__wasm__` still needed before the `raise` call? Does it actually provide an implementation for `raise`? :)

jedisct1:
It's actually optional, and requires an additional library to be linked in order to be emulated. But emulation is very limited.

PR #1106, opened 2021-09-22
Add library for Macos-arm64

SeanMollet:
Exactly what it sounds like. I use sodium core on an Apple Silicon mac and with .net 6 supporting macos-arm64, this library needed to be built for it.

Currently, github doesn't offer Apple Silicon build hardware, so I'm cross compiling on macos-x64. The tests obviously won't work since the library is for arm64 and the x64 mac can't run it.


SeanMollet:
The tests obviously won't work in the CI system since it's cross compiled and the macos-x64 host can use the arm64 library. I ran the tests on my real Apple Silicon HW and they all pass.

<pre>
PASS: aead_aes256gcm
PASS: aead_aes256gcm2
PASS: aead_chacha20poly1305
PASS: aead_chacha20poly13052
PASS: aead_xchacha20poly1305
PASS: auth
PASS: auth2
PASS: auth3
PASS: auth5
PASS: auth6
PASS: auth7
PASS: box
PASS: box2
PASS: box7
PASS: box8
PASS: box_easy
PASS: box_easy2
PASS: box_seal
PASS: box_seed
PASS: chacha20
PASS: codecs
PASS: core1
PASS: core2
PASS: core3
PASS: core4
PASS: core5
PASS: core6
PASS: ed25519_convert
PASS: generichash
PASS: generichash2
PASS: generichash3
PASS: hash
PASS: hash3
PASS: kdf
PASS: keygen
PASS: kx
PASS: metamorphic
PASS: misuse
PASS: onetimeauth
PASS: onetimeauth2
PASS: onetimeauth7
PASS: pwhash_argon2i
PASS: pwhash_argon2id
PASS: randombytes
PASS: scalarmult
PASS: scalarmult2
PASS: scalarmult5
PASS: scalarmult6
PASS: scalarmult7
PASS: scalarmult8
PASS: secretbox
PASS: secretbox2
PASS: secretbox7
PASS: secretbox8
PASS: secretbox_easy
PASS: secretbox_easy2
PASS: secretstream
PASS: shorthash
PASS: sign
PASS: sodium_core
PASS: sodium_utils
PASS: sodium_version
PASS: stream
PASS: stream2
PASS: stream3
PASS: stream4
PASS: verify1
PASS: sodium_utils2
PASS: sodium_utils3
PASS: core_ed25519
PASS: core_ristretto255
PASS: pwhash_scrypt
PASS: pwhash_scrypt_ll
PASS: scalarmult_ed25519
PASS: scalarmult_ristretto255
PASS: siphashx24
PASS: xchacha20
============================================================================
Testsuite summary for libsodium 1.0.18
============================================================================
# TOTAL: 77
# PASS:  77
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[2]: Nothing to be done for `check-am'.
make[1]: Nothing to be done for `check-am'.
</pre>

ektrah:
This is great; thank you very much!

I have two smaller concerns/questions:

1. The target architecture is pretty specific: `aarch64-apple-darwin20.6.0`. I guess this might need to be updated at some point in the future. Who will be maintaining this? Is there maybe an alternative way to do this that requires less maintenance?

2. Without running automated tests at all, we might be shipping a binary that doesn't actually work. I'd feel much more comfortable if `make test` would be run and there was a basic test of the nuget package at the end, like in [#1099](https://github.com/jedisct1/libsodium/pull/1099). Do you think this could be done in some way?

cc @enclave-alistair

SeanMollet:
@ektrah 

1. aarch64-apple-darwin-20.6.0 matches the version of MacOS github makes available under the macos-11 image. The resulting library should work for at least the next few MacOS major releases. When/if there's a breaking upgrade or Github adds M1 build hardware, I'll gladly update it (since I'm going to continue to need this support)

2. I'm not following what you're asking. MacOS arm can't be run in QEMU and Github's CI infrastructure doesn't have real Macos-arm hardware available. This package is a bundle of native libraries, testing them requires either an emulator or real hardware.

SeanMollet:
@ektrah There is no way that I'm aware of to test a macosx-arm64 native library on github actions (they do not have suitable hardware and there is no emulator). Do you have a proposal for how to do that?

Or, is your intention to block this PR until such time as github adds apple silicon CI hardware?

SeanMollet:
I've rebased against the current next.

SeanMollet:
@jedisct1 Yes, a fat binary would be better, but that's not how .net 6 is doing it. They're keeping the platforms separate in order to continue supporting previous versions of .net that don't have native arm binaries. 

I don't know all of the factors involved in their decision to solve it this way, just that this is how they did it. For reference, here's one of my PRs to the dotnet runtime, adding apple silicon support for another native library. Note that it's packaged the same way that I did here.

https://github.com/dotnet/runtime/pull/60310

PR #1130, opened 2021-11-16
Enable building win/arm64 package using visual studio

nsait-linaro:
This patch enables building packages for windows on arm64 using visual studio.

As ARM64 build tools are not available with VS 2015 and v140 toolchain the solution has been updated to target v142.


nsait-linaro:
> Hi!
> 
> And thanks for help with this!
> 
> Is there a way to cross-compile to arm64 using the Community version of Visual Studio?
> 
> Even with MSVC 2022, my tentative attempts miserably failed, but you know the topic way better than I do.
> 
> Is there a way to test this?
> 
> The solution in the root folder is mostly historical, and only used for AppVeyor and running the test suite.
> 
> Actual packages are made with the scripts from `builds/msvc/build` and solutions from `builds/msvc/vs*`
> 
> These are all automatically generated from templates by the `regen-msvc/regen-msvc.py` script.

Thanks, @jedisct1 for the quick response.

Yes, I think it is available in the community version also. You will have to add the following components during installation or launch visual studio installer and modify components

- MSVC v142 - VS 2019 C++ ARM64 build tools (latest)
- C++ ATL for latest v142 build tools (ARM64)

Unfortunately, there is no public VM available for win/arm64 yet in the cloud. Hopefully very soon Azure and others would bring up instances we could use for testing. 

Thanks for explaining the build flow. I will update the auto-generation scripts as well.

jedisct1:
Thank you so much, Niyas!

nsait-linaro:
Thanks @jedisct1 for review and for accepting the patch

PR #1138, opened 2021-12-28
update android build scripts to support latest ndk version and prebuilt toolchains

ReenigneCA:
I'm not sure if it's desired to maybe create a second set of scripts instead. The existing android build scripts don't work with the latest ndk. The latest ndk dropped support for arm prior to armv7a and mips. It is now recommended that prebuilt toolchains be used instead of creating a toolchain for each build (this also saves a lot of disk space.) I've tested this new script and it works on the latest development version of Ubuntu and the latest version of MacOS. perhaps a different folder with android-legacy or something like that for the old architectures?

### DIFF
--- a/dist-build/Makefile.am
+++ b/dist-build/Makefile.am
@@ -1,11 +1,8 @@
 
 EXTRA_DIST = \
 	android-build.sh \
-	android-arm.sh \
 	android-armv7-a.sh \
 	android-armv8-a.sh \
-	android-mips32.sh \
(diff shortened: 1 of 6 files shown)


jedisct1:
Thanks!

But why were the mips versions deleted? Is mips now an unsupported target?

jedisct1:
> The latest ndk dropped support for arm prior to armv7a and mips

Scratch that, I overlooked that part of the description :)

PR #1140, opened 2021-12-29
convenience android multiarch compilation

ReenigneCA:
After a good deal of experimenting I have revised the scripts I submitted for https://github.com/jedisct1/libsodium/pull/1139 such that they generate a directory structure that is easily useable with CMAKE. I confused myself duplicating structures I'd seen other places and then thinking some of my libraries defaulted to that structure... It turns out that CMAKE will search in subdirectories based on the name of the compiler used, with the exception of armv7a where it requires the directory to be named arm-linux-androideabi. setting a PREFIX in android.sh and running it will result in a directory structure like:
```
${PREFIX/include}
sodium
sodium.h


${PREFIX}/lib
aarch64-linux-android:
libsodium.a  libsodium.la  libsodium.so  pkgconfig

arm-linux-androideabi:
libsodium.a  libsodium.la  libsodium.so  pkgconfig

i686-linux-android:
libsodium.a  libsodium.la  libsodium.so  pkgconfig

x86_64-linux-android:
libsodium.a  libsodium.la  libsodium.so  pkgconfig
```

This structure can then be used for all 4 architectures using CMAKE via:
```
list(APPEND CMAKE_FIND_ROOT_PATH "${PREFIX}")
find_library(libsodium NAMES sodium REQUIRED)
```
and adding libsodium as a target_link_library. 

The android.sh build script now echos instructions on how to do this after it runs. That also displays an alternate form for static linking which requires logic for x86. x86 static linking won't work with the current compiles because of text relocation rules. (I tried naively setting -fPIC in the CFLAGS for x86 but it appears that the problem is likely with assembly files. https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md )

I also cleaned up the various variables. The ARCH variable wasn't used at all so I removed it. I left the march flag hard coded in the scripts for each specific architecture as it wasn't meaningful outside that script. TARGET_ARCH wasn't necessary at all. I added a TARGET_ARCH_DIR variable which is normally the same as HOST_COMPILER except for armv7a. Additionally I added a commented out line in android.sh to make it clear how to disable the minimal build which is the default (the original scripts had minimal as the default.)

I chose to target CMAKE as that's what Android Studio uses and I think this layout optimizes for simple compiling. I like to keep my android libs separate from my main ones but I believe setting /usr/local as the PREFIX would mean this works without needing to add the PREFIX to the CMAKE_FIND_ROOT_PATH.


dist-build/android-build.sh
```diff
@@ -1,11 +1,11 @@
 #! /bin/sh
 
 if [ -z "$NDK_PLATFORM" ]; then
-  export NDK_PLATFORM="android-19"
+  export NDK_PLATFORM="android-16"
```
jedisct1, line 4: Why revert to `android-16`? Looks like the lowest version supported by the NDK is now `19`.

dist-build/android-armv7-a.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
-export TARGET_ARCH=armv7-a
-export CFLAGS="-Os -mfloat-abi=softfp -mfpu=vfpv3-d16 -mthumb -marm -march=${TARGET_ARCH}"
-ARCH=arm HOST_COMPILER=armv7a-linux-androideabi "$(dirname "$0")/android-build.sh"
+export CFLAGS="-Os -mfloat-abi=softfp -mfpu=vfpv3-d16 -mthumb -marm -march=armv7a"
+HOST_COMPILER=armv7a-linux-androideabi TARGET_ARCH_DIR=arm-linux-androideabi "$(dirname "$0")/android-build.sh"
```
jedisct1, line 3: `HOST_COMPILER` is not very self-explanatory. One may expect to set it to something such as `clang` or `zig cc`. Maybe something like `NDK_COMPILER_DIR`?

dist-build/android-build.sh
```diff
@@ -18,18 +18,23 @@ if [ ! -f ./configure ]; then
...
+if [ -z ${INCLUDE_PATH+x} ]; then export INCLUDE_PATH="${PREFIX}/include"; fi
+if [ -z ${TARGET_ARCH_DIR+x} ]; then export TARGET_ARCH_DIR="${HOST_COMPILER}"; fi
+
+export TOOLCHAIN_OS_DIR="`uname | tr '[:upper:]' '[:lower:]'`-x86_64/"
 export TOOLCHAIN_DIR="$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${TOOLCHAIN_OS_DIR}"
 echo "$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/${TOOLCHAIN_OS_DIR}/${HOST_COMPILER}"
 
 export PATH="${PATH}:${TOOLCHAIN_DIR}/bin"
-SDK_VERSION_NUM=$(echo $NDK_PLATFORM | cut -d'-' -f2)
+SDK_VERSION_NUM=`echo $NDK_PLATFORM | cut -d'-' -f2`
```
jedisct1, line 37: Modern syntax is preferred. I'd highly recommend using the `shellcheck` tool to check the script against common errors, including obsolete constructions. There's also a Visual Studio Code plugin for it. It's a great tool!

dist-build/android-build.sh
```diff
@@ -71,9 +78,11 @@ if [ "$NDK_PLATFORM" != "$NDK_PLATFORM_COMPAT" ]; then
     ${LIBSODIUM_ENABLE_MINIMAL_FLAG} \
     --host="${HOST_COMPILER}" \
     --prefix="${PREFIX}" \
+    --includedir="${INCLUDE_PATH}" \
+    --libdir="${LIB_PATH}" \
     --with-sysroot="${TOOLCHAIN_DIR}/sysroot" || exit 1
 
-  grep -E '^#define ' config.log | sort -u >config-def.log
+  egrep '^#define ' config.log | sort -u >config-def.log
```
jedisct1, line 85: `grep -E` is more portable.

dist-build/android.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/sh
+
+# uncomment below to disable the minimal build
```
jedisct1, line 3: Distributed files are not supposed to be modified. Maybe allow these parameters to be set on the command-line instead?

dist-build/android.sh
```diff
@@ -0,0 +1,44 @@
...
+#export LIBSODIUM_FULL_BUILD=TRUE
+# uncomment below and change the install location(s) to where you keep your ndk libraries and includes
```
jedisct1, line 5: The comment is a bit misleading. I initially thought it represented the path where the NDK was installed.

Is the install prefix necessary? Android libraries will not run on the host operating system, so installing them in `/usr/local` sounds a bit strange. Is there anything wrong with installing the files locally, and then people will move them wherever they want?

dist-build/android-x86_64.sh
```diff
@@ -1,4 +1,3 @@
 #!/bin/sh
-export TARGET_ARCH=westmere
-export CFLAGS="-Os -march=${TARGET_ARCH}"
-NDK_PLATFORM=android-21 ARCH=x86_64 HOST_COMPILER=x86_64-linux-android "$(dirname "$0")/android-build.sh"
+export CFLAGS="-Os -march=westmere"
+NDK_PLATFORM=android-21 HOST_COMPILER=x86_64-linux-android TARGET_ARCH_DIR="${HOST_COMPILER}" "$(dirname "$0")/android-build.sh"
```
jedisct1, line 3: Should `NDK_PLATFORM` be target-specific? Maybe it can be set only once in `android-build.sh`.

dist-build/android-build.sh
```diff
@@ -18,18 +18,23 @@ if [ ! -f ./configure ]; then
...
   echo "You shouldn't use android-build.sh directly, use android-[arch].sh instead" >&2
   exit 1
 fi
 
-export PREFIX="$(pwd)/libsodium-android-${TARGET_ARCH}"
-export TOOLCHAIN_OS_DIR="$(uname | tr '[:upper:]' '[:lower:]')-x86_64/"
+if [ -z ${PREFIX+x} ]; then export PREFIX="$(pwd)/libsodium-android"; echo "setting default PREFIX=${PREFIX}"; fi
+if [ -z ${LIB_PATH_PREFIX+x} ]; then echo ""; else export LIB_PATH="${LIB_PATH_PREFIX}/${TARGET_ARCH_DIR}"; fi
+if [ -z ${LIB_PATH+x} ]; then export LIB_PATH="${PREFIX}/lib/${TARGET_ARCH_DIR}"; fi
+if [ -z ${INCLUDE_PATH+x} ]; then export INCLUDE_PATH="${PREFIX}/include"; fi
```
jedisct1, line 29: No need for `if` statements. Use `export INCLUDE_PATH="${INCLUDE_PATH:-${PREFIX}/include}"`.

jedisct1, line 29: Does `cmake` really have an issue with header files being in `${PREFIX}/${ARCH}/include` and libs being in `${PREFIX}/${ARCH}/lib`?

Having different prefixes makes the layout different from system libraries. It also assumes that header files are exactly the same for all targets, which may not always be the case.

dist-build/android-build.sh
ReenigneCA: I didn't notice this. I think it was 16 in the stable version.

dist-build/android-armv7-a.sh
ReenigneCA: I'm happy to change it the variable name comes from the version before my edits. it's HOST_COMPILER because it's the name of the compiler on the host that's used for the cross compilation. It's actually a program so NDK_COMPILER would make more sense than NDK_COMPILER_DIR.

dist-build/android-build.sh
ReenigneCA: the if statements allow someone to override the defaults to suit them. in android.sh there are commented out lines that someone can modify to do something custom without reaching into the scripts.

dist-build/android-x86_64.sh
ReenigneCA: I have no issues making it target specific. I left it as per the old scripts I figured the intention was to maximize legacy support but I never really researched it.

dist-build/android.sh
ReenigneCA: I'm happy to make that change when I have time. The old scripts didn't have any easy way to configure them and this just seemed like the fastest way.

dist-build/android-build.sh
ReenigneCA: sorry again I didn't actually change this line another lost change from stable.

dist-build/android.sh
ReenigneCA: It does sound strange to me I misunderstood a previous comment of yours and thought this was the approach you preferred for cross compilation ie to match make install's layout.  The prefix can be set to anything when I use it I set it to /aSharedDrive/ndk-root/ I think it actually does work with /usr/local because all of the libs are inside folders that specify the architecture and operating system.

dist-build/android-build.sh
ReenigneCA: This is related to me thinking matching something similar to /usr/lib/x86_64-linux-gnu was your desired approach to simplify cross compilation. It definitely can work I think it requires using cmake variables instead of just specifying a root folder.

dist-build/android-build.sh
ReenigneCA: sorry never heard of shellcheck I'll run my scripts through it before any more commits.

ReenigneCA:
So to sum up where I'm at:

I forgot to take your updates to my previous commit sorry for all the little syntax things that seeped back in.

Some of the variable names and values I just maintained from the old version (some erroneously from the stable tarball,) I'll clean them up to the current values from master or in the case of variable names I'll change them when I get a thumbs up on the appropriate comment.

I mostly write code and actually really hate dealing with build systems and especially with shell scripts so I misunderstood the desired output format. The old format had the architecture embedded in the prefix itself so you couldn't easily have multiple architectures and multiple libraries in the same directory structure. As long as the arch is there this shouldn't be a problem. I'll switch to a ${PREFIX}/${ARCH}/include ${PREFIX}/${ARCH}/lib format and clean up the CMAKE recommendations. I'm pretty sure the only real difference CMAKE wise will be the need to specify the arch when setting the search root. I have to run for now I'll go through and resolve comments when I get a chance and alter the default structure after I get a chance to test the differences with an Android project.

dist-build/android-build.sh
ReenigneCA: I did some experimenting. As far as I know you have to specify include directories no matter how they are organized cmake won't search for folders named include or deal with architecture variables so whether it's ${PREFIX}/${ARCH}/include or or ${PREFIX}/include/${ARCH}/ it doesn't make a difference as far as I can see. For libraries find_library will search for architectures inside lib folders but not the other way around. This is easy to work around you just have to specify a different root path to search based on the same variable you need for includes anyway list(APPEND CMAKE_FIND_ROOT_PATH "${PREFIX}/${CMAKE_LIBRARY_ARCHITECTURE}")

Thinking about it a bit more and doing some more research I think I still prefer the lib/${ARCH} approach. It does match the way linux handles libraries in the /usr/lib folder as well as the way platform specific includes are handled in /usr/include. The ndk compiler for aarch64 has a default search path for includes:  $ANDROID_NDK_HOME/toolchains/llvm/prebuilt/linux-x86_64/bin/./../sysroot/usr/include/aarch64-linux-android 
it actually doesn't have one for the sysroot/usr/local/include. Researching that a bit more I now realize my concept of /usr/local/lib has been wrong my whole life. I've always put anything I compiled myself in /usr/local/ but even if I do so I should have been putting libraries in /usr/lib. So anyway if we do ${PREFIX}/lib/${ARCH} then someone could set the prefix to their ndk sysroot and then the appropriate architecture's ndk compiler will just find the includes and the libs with no need for any special handling by CMAKE or any other build tool. That option isn't available if we use ${ARCH}/lib .

thinking a bit more about your comment about matching other architectures I was thinking about the host architecture not other scripts in dist-build. they all store their files in libsodium-${arch} folders I don't think this approach really deviates as that folder is just the default $PREFIX the script as it currently exists will put all the files in libsodium-android and then have the lib/include $ARCH stuff within that folder. I don't think any of the other dist-build scripts support more than one architecture. For the automated build infrastructure this would mean the resulting files always have exactly the same names as the old scripts used to include the android version though I discuss that in the para below. It sort of touches on a different comment you made about android versioning and where the version should be set but I discuss it here as it's relevant to the preferred directory layout as well.

The only remaining weirdness for Android is the different android versions. The ndk compilers and cmake don't take these into account when looking for folders. I sort of expect that most people just compile all their ndk libs for the lowest common denominator and rely on forwards compatibility. If someone wants to have different versions of libraries that support different android versions they'll need to use some sort of build system to manage that no matter what AFAIK. I misread the original scripts I thought it was falling back to a compat version if there was a failure but it actually originally compiled a specified version (based on script variables) and then it would compile a compat version as well unless the compat version and specified version were the same. When I updated the scripts now it will actually compile the compat version and replace the version that was asked for. There aren't any useful architecture based names that include android version (ie ones that clang or cmake will find automatically) so the best way to deal with this would be to just change the prefix by appending the android version. This would of course require special handling if a custom prefix was specified. We don't want "sysroot-android21" in an ndk if anyone actually does want to use their sysroot so if a PREFIX is specified we shouldn't do the compat check... but if the compat version is the default then that means the code would never be called... I don't think this compat fallback actually makes any sense as the old scripts had to be edited to change the target version anyway so why not just have the compat version the default and if the user wants to specify something different we should trust them and not compile a second compat version.

dist-build/android-x86_64.sh
ReenigneCA: I'm going to do one commit after I get everything sorted as some of the items might no longer be relevant once the larger layout is decided so I'm going to mark this and a few other minor items as resolved but no code change will be visible yet.

dist-build/android-armv7-a.sh
ReenigneCA: NDK_COMPILER also isn't technically correct as it's only part of the compiler's name NDK_COMPILER_PREFIX might make more sense though it might also be a different kind of confusing... My preference is still with NDK_COMPILER but I don't have strong feelings about this item and will go with whatever you're happy with.

dist-build/android.sh
ReenigneCA: Can you clarify how you'd prefer this? command line argument? prompt after running the script? 

If I use command line arguments and there are defaults many people will not realise what defaults are being set which might bite them if they use functionality that isn't included in a minimal build (though that default could be changed). With no defaults the scripts of course won't do anything unless arguments are specified. What are your thoughts on using command line arguments if specified and if none are specified printing out what the defaults are and giving a Y/N prompt? Without a prompt any messaging will fly by as the configure scripts get run and the compilations happen. The other option is prompting for different options on stdin after the program is run or potentially only doing this if options are missing though this can be painful for anyone running the script from a script.

I guess my gut is how will anyone understand the commandline arguments without looking at the script or not having defaults and having the script complain if nothing is provided or having a system of prompts to confirm anything that isn't provided. If someone does look at the script isn't setting the variables in the script easier for everyone? 

So how about I go with script variables for now and take on a future task to make a more intelligent command line argument system that uses flags and has prompts if anything is missing?

dist-build/android.sh
ReenigneCA: this makes more sense in the context of my reply about the directory structure layout but how about I change the example prefix to "${ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/$(uname | tr '[:upper:]' '[:lower:]')-x86_64/sysroot" and add another one to /custom/build/location ?

jedisct1:
A `pkg-config` configuration is also installed, so maybe we can recommend using it instead.

This has several advantages:

- This is independent of where headers and libraries are located; will also work with a local directory rather than a system install
- This works out of the box with most build systems, including automake, bazel, zig, and meson
- It also takes care of additional compilation/link flags that may be required for proper operation
- It may be more future-proof.

I don't know much about `cmake`, but it seems to have out of the box support for `png-config` as well:

```text
find_package(PkgConfig REQUIRED)
pkg_search_module(LIBSODIUM REQUIRED libsodium>=1.0.18)

target_compile_options(HelloWorld PUBLIC ${LIBSODIUM_STATIC_CFLAGS} ${LIBSODIUM_STATIC_CFLAGS_OTHER} ${LIBSODIUM_STATIC_LDFLAGS} ${LIBSODIUM_STATIC_LDFLAGS_OTHER})
target_include_directories(HelloWorld PUBLIC ${LIBSODIUM_STATIC_INCLUDE_DIRS})
target_link_directories(HelloWorld PUBLIC ${LIBSODIUM_STATIC_LIBRARY_DIRS})
target_link_libraries(HelloWorld PUBLIC ${LIBSODIUM_STATIC_LIBRARIES})
```

And the base directory just has to be specified in `CMAKE_PREFIX_PATH`.

Regarding the target API version, libsodium doesn't use much of the native Android system libraries, so targeting the lowest supported API may be fine. Unless an application designed for a different API won't be able to link a library compiled for a lower API?
If that is the case, we may want to include the API level in the base directory name.

jedisct1:
There was also a subtle difference between `TARGET_ARCH` and the architecture as exposed in directory names.

The target architecture can include CPU feature flags (such as `+crypto` for `armv8-a`). The resulting library will work on the baseline CPU without these extensions, however, if these extensions are detected at compile-time, they will be enabled. And in order for the compiler to accept code using these extensions, they have to be enabled at compile-time.

This is what broke the ARM builds in that PR.

ReenigneCA:
After all of my investigation into the best way to lay out libraries for the ndk I've come full circle. The way I was trying to lay things out is according to the standard the ndk uses for it's sysroot (I actually also discovered that you can make separate directories for different android versions it uses the layout that this pull request uses plus a directory that's just the android api number,) and if you put things in that format into the sysroot  lib folder things compile using -lsodium. This seemed to me to be a very convenient way to set things up, but while programs will compile, android studio will not include the libs in a generated apk if they are in the sysroot, which in retrospect makes a lot of sense. Personally I like this format better still only because I want one lib folder with all my archs in it, but that is definitely outweighed by anything that depends on the current format and is something I can do with a python script on my end. So while I really thought I was learning things to make things better for everyone after everything I now see I was on the totally wrong track. 

That being said this rabbit hole has raised a few questions about the current scripts

- rename HOST_COMPILER variable? NDK_COMPILER?;
- add a comment about TARGET_ARCH and how it's used (how exactly is it used? it isn't passed to anything I guess the make process grabs it from env?) so anyone who updates the scripts in the future knows to leave it alone;
- the current script has configuration parameters that someone could tweak if they want to, given I now see no reason to alter the directory layout or make it configurable and +crypto is on by default (I can't see why anyone would ever turn that off,) the only thing that I think might merit a prompt or commandline argument is the --enable-minimal flag which is on be default in all of the dist-build scripts it appears. It's shut off in all of them if $LIBSODIUM_FULL_BUILD has a value is this variable set somewhere else somehow?  I haven't really seen any info anywhere on the intended way to build libsodium for use with the ndk just how to use libraries that are built on top of it to make it available to java/kotlin. I believe the answer is to download the source and run the dist-build android scripts directly which will result in minimal builds and require someone to sort through the scripts to get a full build if they require it.   
- right now armv8a and x86_64 specify NDK_PLATFORM 21 while the others fall through to the default in android-build.sh and get set to 19. I'm not sure what the significance of these versions is. [The current supported LTS of the ndk is 23](https://developer.android.com/ndk/downloads) is there any reason not to set this as the default in android-build.sh and remove other references?
- The scripts use an NDK_PLATFORM_COMPAT variable to potentially compile a second version, but it's unclear to me how/where that's supposed to be set by a user? with the existing naming convention the functionality does make sense but is it actually used anywhere and how should it be made available?
- could still use an android.sh script that builds each of the architectures. Depending on what is desired for the differing NDK_PLATFORM variables it could set NDK_PLATFORM and the arch specific scripts could check if it has a value and only set if it doesn't. Perhaps the other scripts can generally be left alone and android.sh could provide an interface more suitable to a new user with prompts and such, effectively a front end to the android build system. 

I think it makes sense to close this pull request without merging it but if you want to comment on / add to the above list I'll make a new branch (not from an old stable version this time I promise :P ) and try to cover everything off.

PR #1188, opened 2022-06-14
RFC9381 ECVRF implementation

iquerejeta:
Implementation of [RFC9381](https://datatracker.ietf.org/doc/html/rfc9381). 

Made a change to ed25519_ref10 for performance reasons. Mainly, I have included a variable base variable time scalar multiplication (to compute `a * U + b * V` for variable `U` and `V`). 

Implements #1177


test/default/vrf.c
```diff
@@ -0,0 +1,135 @@
...
+            "c5aa8df43f9f837bedb7442f31dcb7b166d38535076f094b85ce3a2e0b4458f7",
+            "fc51cd8e6218a1a38da47ed00230f0580816ed13ba3303ac5deb911548908025",
+            "926e895d308f5e328e7aa159c06eddbe56d06846abf5d98c2512235eaa57fdce35b46edfc655bc828d44ad09d1150f31374e7ef73027e14760d42e77341fe05467bb286cc2c9d7fde29120a0b2320d04",
+            "121b7f9b9aaaa29099fc04a94ba52784d44eac976dd1a3cca458733be5cd090a7b5fbd148444f17f8daf1fb55cb04b1ae85a626e30a54b4b0f8abf4a43314a58",
+        }
+};
+
+static const unsigned char messages[3][2] = {{0x00}, {0x72}, {0xaf, 0x82}};
+
+static inline void printhex(const char *label, const unsigned char *c, size_t len){
```
jedisct1, line 37: We have `sodium_bin2hex()`.

test/default/vrf
```diff
@@ -0,0 +1,210 @@
+#! /bin/sh
```
jedisct1, line 1: This file is automatically generated at compile time, it should not be imported in git.

src/libsodium/crypto_vrf/ietfdraft12/keypair.c
```diff
@@ -0,0 +1,40 @@
...
+
+#include "crypto_hash_sha512.h"
+#include "crypto_vrf_ietfdraft12.h"
+#include "private/ed25519_ref10.h"
+#include "randombytes.h"
+#include "utils.h"
+
+int
+crypto_vrf_ietfdraft12_seed_keypair(unsigned char *pk, unsigned char *sk,
+                                 const unsigned char *seed)
```
jedisct1, line 11: Indentation is weird here :)

src/libsodium/crypto_vrf/ietfdraft12/vrf_ietfdraft12.h
```diff
@@ -0,0 +1,10 @@
+#ifndef vrf_ietfdraft12_H
+#define vrf_ietfdraft12_H
+
+static const unsigned char SUITE = 0x04; /* ECVRF-ED25519-SHA512-ELL2 */
+
+static const unsigned char ZERO = 0x00;
```
jedisct1, line 6: Just use 0, 2 and 3 :)

test/default/vrf.c
```diff
@@ -0,0 +1,135 @@
...
+
+int main(void)
+{
+    unsigned char *seed, *expected_pk, *expected_proof, *expected_output;
+    seed            = (unsigned char *) sodium_malloc(crypto_vrf_ietfdraft12_SEEDBYTES);
+    expected_pk     = (unsigned char *) sodium_malloc(crypto_vrf_ietfdraft12_PUBLICKEYBYTES);
+    expected_proof  = (unsigned char *) sodium_malloc(crypto_vrf_ietfdraft12_BYTES);
+    expected_output = (unsigned char *) sodium_malloc(crypto_vrf_ietfdraft12_OUTPUTBYTES);
+
+    unsigned char sk[64];
```
jedisct1, line 54: libsodium uses C89, since some supported compilers still don't support C99 yet.

Would you mind moving all the declarations to the top of the block?

test/default/vrf.c
```diff
@@ -0,0 +1,135 @@
...
+    const char seed[2 * 32 + 1];
+    const char pk[2 * 32 + 1];
+    const char proof[2 * 80 + 1];
+    const char output[2 * 64 + 1];
+} TestData;
+/*
+ * Test data taken from https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vrf-09#appendix-A.4
+ */
+static const TestData test_data[] = {
+        {
```
jedisct1, line 15: Indentation is weird :)

src/libsodium/crypto_vrf/ietfdraft12/verify.c
```diff
@@ -0,0 +1,117 @@
...
+    crypto_hash_sha512_update(&hs, &TWO, 1);
+    crypto_hash_sha512_update(&hs, Y_string, 32);
+    crypto_hash_sha512_update(&hs, H_string, 32);
+    crypto_hash_sha512_update(&hs, pi, 32);
+    crypto_hash_sha512_update(&hs, U_string, 32);
+    crypto_hash_sha512_update(&hs, V_string, 32);
+    crypto_hash_sha512_update(&hs, &ZERO, 1);
+    crypto_hash_sha512_final(&hs, hram);
+
+    return crypto_verify_16(c, hram);
```
jedisct1, line 101: For ed25519, `hram` meant `H(R || A || M)`.

It doesn't match what we are doing here.

src/libsodium/crypto_vrf/ietfdraft12/vrf_ietfdraft12.h
jedisct1: Nevermind. Them being uppercase made me think they were macros, and not variables you need to use the address of.

src/libsodium/include/sodium/private/ed25519_ref10.h
```diff
@@ -99,6 +99,12 @@ void ge25519_double_scalarmult_vartime(ge25519_p2 *r, const unsigned char *a,
                                        const ge25519_p3 *A,
                                        const unsigned char *b);
 
+void
+ge25519_double_scalarmult_vartime_variable(ge25519_p2 *r, const unsigned char *a,
```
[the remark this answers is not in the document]
iquerejeta: I was thinking that instead of checking if the second input is the generator, we could have an additional argument, `unsigned int b`, that determines whether we are using the generator as a second parameter point or not. If yes, then that second ge25519_p3 would be any dummy variable with that type, as it would be ignored by the function. However, that might be quite confusing for readability. Maybe we could do this if we still expose both functions, but they internally use the same underlying function with this conditional argument.

Otherwise, we could include the generator as a constant, and force the caller of `double_scalarmult_vartime` to give the generator as an input. This, together with a new function `is_generator()` would also work. 

What direction you think is best?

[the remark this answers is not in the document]
jedisct1: Couldn't `NULL` represent the generator?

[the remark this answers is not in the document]
iquerejeta: Oh, of course, yes, we could pass `NULL` as the second point.

test/default/vrf.c
iquerejeta: I've changed the test styles, so that matching verification happens with `vrf.exp`, following the style of hashing tests.

thecontemoraryflorals (approved):

thecontemoraryflorals (approved):

iquerejeta:
@jedisct1 , does this mean that we are good to push this forward? If yes, I'll make a pass to make sure that the implemented version is what got finalised in the draft, and mark it as ready.

jedisct1:
@iquerejeta Yes :)

iquerejeta:
Got some trouble reproducing the `undefined reference to 'crypto_core_ed25519_scalar_negate'` and `undefined reference to 'crypto_core_ed25519_scalar_negate'` errors locally. I made sure I was configuring with the same call as the CI action: 
```
env CPPFLAGS="-DDEV_MODE=1" ./configure --disable-dependency-tracking --enable-minimal
```

But couldn't reproduce it.

iquerejeta:
So, I've managed to reproduce them, but no progress in debugging. I'll try again tomorrow. Any idea why we might have 

```
Undefined symbols for architecture arm64:
...
ld: symbol(s) not found for architecture arm64
``` 

type errors for symbols that are exposed in the public API when called within the library?

jedisct1:
Errrrr... this is super weird.

How did you manage to reproduce it? Is it non-deterministic?

jedisct1:
Actually it's not weird. 

The `core_ed25519` functions are not available in MINIMAL builds.

iquerejeta:
Ok, makes sense. What is your preferred way forward? To have `core_ed25519` exposed in the MINIMAL build, or instead have `from_string` and `scalar_negate` as part of `ed25519_ref10.c`? Seeing how the library is organised, I presume the second, but just want to check with you before making further changes.

jedisct1:
Yeah, I agree that the second option looks better.

nurupo:
Look like this is the first time a VLA was used in libsodium. At least according to ComCert, a formally verified mostly-C99 (it doesn't support VLAs) compiler intended for life-critical and mission-critical software:

```
libsodium/src/libsodium/crypto_vrf/rfc9381/verify.c:50: error: size of array is not a compile-time constant
```

Not sure if you have a no-VLA policy or anything like that, just thought I would let you know.

jedisct1:
VLAs and `alloca` should not be used, as it's hard to know what the usage limits are before getting a stack overflow. And as documented, CompCert is one of the compilers libsodium is always tested with. That code hasn't been reviewed yet, but that's one of the things that will be changed before it gets merged into `stable`.

PR #1210, opened 2022-09-19
Remove the support for bitcode [ios]

jnavarrom:
According to Xcode 14 documentation:

> Deprecations
> Starting with Xcode 14, bitcode is no longer required for watchOS and tvOS applications, and the App Store no longer accepts bitcode submissions from Xcode 14.
> 
> Xcode no longer builds bitcode by default and generates a warning message if a project explicitly enables bitcode: “Building with bitcode is deprecated. Please update your project and/or target settings to disable bitcode.” The capability to build with bitcode will be removed in a future Xcode release. IPAs that contain bitcode will have the bitcode stripped before being submitted to the App Store. Debug symbols for past bitcode submissions remain available for download. (86118779)

> Because bitcode is now deprecated, builds for iOS, tvOS, and watchOS no longer include bitcode by default. (87590506)

https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-14-release-notes

### DIFF
--- a/dist-build/apple-xcframework.sh
+++ b/dist-build/apple-xcframework.sh
@@ -87,8 +87,8 @@ build_ios() {
   export SDK="${BASEDIR}/SDKs/iPhoneOS.sdk"
 
   ## 32-bit iOS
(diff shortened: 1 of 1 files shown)


jedisct1:
Yeah. I'm wondering if this is too early or if this is something we should already commit to `stable`.

PR #1211, opened 2022-09-23
Add support for arm64 watchOS

Elih96:
According to the Xcode 14 release notes:

> Xcode builds for watchOS devices now include the arm64 architecture by default. (83319300)

[https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes)

As a result adding swift-sodium to a watchOS target fails during the Archive step due to missing arm64 architecture.

Here's how I validated - I ran ./apple-xcframework.sh, switched out the xcframework in swift-sodium for this one, tested for watchOS, successfully built.


jedisct1:
Thank you!

PR #1224, opened 2022-11-12
Add proof-of-concept for AES-256-GCM on ARM, not ready yet

pgmcgee:


pgmcgee:
Sorry about that, I didn't mean to open it against this repo yet.

jedisct1:
The AES-GCM implementation is soon going to be replaced with one adapted from the Zig standard library.

niXman:
what is Zig?

jedisct1:
> what is Zig?

A modern programming language, whose standard library includes everything libsodium has and much more.

jedisct1:
It currently requires libsodium to be compiled with `clang` or `zig cc`, but the development branch does have support for AES256-GCM on ARM.

That being said, having ARM optimized implementations of pretty much everything else (blake2b, chacha20, salsa20, poly1305) would have been far more interesting.

pgmcgee:
This is great context, thanks. That's exciting that the AES256-GCM algorithm will soon support ARM. 

For the other algorithms, would the sse2neon shim be an appropriate way to convert them or will the ARM intrinsics need to be inline? For AES256-GCM, I was using the shim as a temporary pass and then I was going to convert them over using the shim as a guide. For all those other algorithms, I could probably refactor them to use the shim library, but my knowledge of ARM intrinsics isn't good enough to convert them all at scale.

jedisct1:
`sse2neon.h` would be a hack rather than something we could seriously merge. Emulation would be suboptimal, and Intel intrinsics are not very good as an abstraction layer.

Also, `sse2neon.h` doesn't compile on Visual Studio, so this is a non starter.

pgmcgee:
Yep, that makes total sense, thanks!

PR #1238, opened 2022-12-22
MacCatalyst support

MartyIX:
https://github.com/jedisct1/libsodium/discussions/1235

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -158,44 +158,16 @@ jobs:
           name: build-linux-musl-arm64
           path: zig-out/lib/libsodium.so
 
-  build-macos-x64:
-    runs-on: macos-latest
-    steps:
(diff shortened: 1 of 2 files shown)


jedisct1:
It doesn't really make sense to build the `xcframework` only to use one of the files it generates.

We should compile the `xcframework` once, and use its content for all the Apple targets. And still test them if possible.

ektrah:
@MartyIX Thanks a lot for your pull request and apologies for not responding earlier; things are always going a bit slow after the Christmas break 😌 

I think the best way to test this would be including your test application in this repository. For that, it would be great if you could provide the steps needed to create the test application from scratch in the command line, e.g.:

```bash
$ dotnet new ...
$ dotnet add package libsodium --source path/to/downloaded/github/actions/artifact
... add/modify files ...
$ dotnet build
$ dotnet run
```

PR #1245, opened 2023-01-07
Rename `dotnet-core.yml` to `dotnet.yml`

MartyIX:
This PR makes sure that .NET workflow is picked up by Github Actions in forks of this repo.

Experimentally verified by merging of https://github.com/MartyIX/libsodium/commit/01e21a5dae493b3bed6966adc8864cdd02cc5fef to my fork. See Actions [here](https://github.com/MartyIX/libsodium/actions).

PS: .NET Core is no longer a thing. Basically Microsoft has been trying to unify Mono and .NET Core under the new simple name ".NET". So the change makes sense even from that point of view.


jedisct1:
I don't understand.

What does the name of a GitHub workflow change?

jedisct1:
I still don't get it. Files names for GitHub workflows don't change anything, do they?

Why would a special file name be required for workflows having something to do with .NET?

Maybe `dotnet.yml` is used in examples, but is there actual documentation mentioning that it is required? (so, why do other names still work perfectly fine?)

Unless this is a joke, I'd really like to understand why.

PR #1291, opened 2023-07-19
Add proper CET support (followup to #1290)

jedisct1:
/cc @siddhesh


jedisct1:
Just trying to be a little bit more careful about what's actually usable on the system.

PR #1300, opened 2023-08-26
Allow other zig projects to use libsodium as a dependency

fjebaker:
With the advent of the Zig [Package Manager](https://ziglang.org/download/0.11.0/release-notes.html#Package-Management), I've made a couple of changes to the `build.zig` file that allows libsodium to be used as a dependency in Zig projects.

This includes adding a number of options so that the static and shared library can be built in isolation. On non-Windows OSs, the shared / static artifacts share the same name from Zig's perspective (from my naive understanding), so you get an error message about ambiguity if you try to depend on the artifact. Now, you can just pass some options to avoid that.

The options have been implemented so the default behaviour should be identical to how the `build.zig` script previously worked. An addendum to this is that I have explicitly made Zig include all header files in `zig-out/include` -- again, so downstream projects can easily depend on it.

A downstream library may unambiguously build against the shared or static library by adding the artifact to their `build.zig.zon` and then with e.g.
```zig
const libsodium = b.dependency(
    "libsodium",
    .{
        .target = target,
        .optimize = optimize,
        .@"test" = false, // `test` is a keyword in zig
        .static = false,
    },
);
```
in their `build.zig`.

Hope these changes are okay!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -19,20 +19,34 @@ pub fn build(b: *std.build.Builder) !void {
 
     const enable_benchmarks = b.option(bool, "enable_benchmarks", "Whether tests should be benchmarks.") orelse false;
     const benchmarks_iterations = b.option(u32, "iterations", "Number of iterations for benchmarks.") orelse 200;
+    var build_static = b.option(bool, "static", "Build libsodium as a static library.") orelse true;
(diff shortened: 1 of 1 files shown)


jedisct1:
Thank you! This is great!

PR #1305, opened 2023-09-01
Do not redefine  _mm_roti_epi64 if xop feature set enabled

Scr3amer:
Hello,

Me again with another PR to fix some warnings when compiling the library.
This time it is about macro redefinition.

My toolchain has already _mm_roti_epi64 that returns __builtin_ia32_vprotqi.
However it fails if I don't enable the feature set XOP via -mxop but I am targeting a hardware that doesn't have it so instead I decided to change the code as follow:

- if XOP is enabled (we can check it via "__ XOP __") we don't redefine _mm_roti_epi64
- if XOP is not enabled but the macro is already existing in the toolchain headers, we disable the one coming from the toolchain (since it won't work anyway without XOP) before redefining it in libsodium to avoid preprocessor macro redefinition warnings.

I just have one question: is this redefinition due to security reasons ? There is a flaw with the builtin rotation ? Or it is just to make it work on all targets ? I assumed the latter but I could be wrong, not an expert.

That's all.
Cheers,
Scr3am

PS: Should we put _mm_roti_epi64 in its own header file and include it in the 3 headers ? Instead of copy-pasting it 3 times ?

### DIFF
--- a/src/libsodium/crypto_generichash/blake2b/ref/blake2b-compress-sse41.h
+++ b/src/libsodium/crypto_generichash/blake2b/ref/blake2b-compress-sse41.h
@@ -5,18 +5,23 @@
 #define LOADU(p) _mm_loadu_si128((const __m128i *) (const void *) (p))
 #define STOREU(p, r) _mm_storeu_si128((__m128i *) (void *) (p), r)
 
-#define _mm_roti_epi64(x, c)                                         \
-    (-(c) == 32)                                                     \
-        ? _mm_shuffle_epi32((x), _MM_SHUFFLE(2, 3, 0, 1))            \
-        : (-(c) == 24)                                               \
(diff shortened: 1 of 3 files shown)


jedisct1:
Why not, thanks!

But how did you end up with `_mm_roti_epi64` being defined? 

That requires `<x86intrin.h>` to be included, which we don't do.

What compiler did you use?

Scr3amer:
So this is the include chain I have that leads to x86intrin.h:

. ./include/sodium/private/common.h
.. my_toolchain/lib/clang/16/include/intrin.h
... my_toolchain/include/intrin.h
.... my_toolchain/lib/clang/16/include/x86intrin.h

I use this toolchain https://github.com/mstorsjo/llvm-mingw/releases, version 16.0.6

PR #1309, opened 2023-09-08
Support for arbitrary context in ed25519ph

f4grx:
Hi,

this pr adds support for arbitrary context in ed25519ph. This is required for compatibility with the crypto system being implemented by veilid.

It introduces a new API to set that context, but keep compatibility with all existing API. Not using this API maintains perfect compatibility with the current behaviour. The only side effect is a change to the size of crypto_sign_ed25519ph_state.

I can imagine another solution that does not change the size of crypto_sign_ed25519ph_state but rather introduces two new functions that takes the context as parameter and pass it directly to the underlying functions, so that users can sign and verify with a context. I find this version less elegant, and it would introduce more nesting of internal functions, but I can do it on request.

Please tell me how this pull request can be merged. This is just an initial state, so people can see my intention. I am open to any change.

Thank you.


jedisct1:
Closing, as this is not ready for being merged.

That feature can still be added, provided that there are more use cases.

PR #1332, opened 2023-11-12
Build as many binaries as possible for distribution

js-9:
I needed a bunch of different binaries for a project but couldn't find them from any official source, so I made a github workflow to build as many of them as possible.
Results can be found here: https://github.com/js-9/libsodium/actions/runs/6838395872

Currently builds:
- android (x86, x86_64, armv7a, armv8a)
- linux (i686, x86_64, armv7a, armv8a)
- apple (ios, macos, tvos, watchos, and simulators for all these plus catalyst which I don't even know what it is, plus xcframework for all different architectures and platforms)
- windows (mingw x86 and x64, msvc x86, x64 and arm64 with vs2022)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,181 @@
+name: build all binaries
+on:
+  workflow_dispatch:
+
+jobs:
+  build-android:
+    name: build binaries for android
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x86_64
+          - x86
+          - armv7-a
+          - armv8-a
+    runs-on: ubuntu-latest
+    steps:
+      - name: set output dir name based on arch
+        shell: bash
+        id: set-result-path-part
+        run: |
+          case "${{ matrix.arch }}" in
+            x86_64) RESULT_NAME=westmere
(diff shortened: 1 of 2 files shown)


jedisct1:
Hi!

Proper packages for libsodium are available for available for the vast majority of operating systems and linux distributions. They are guaranteed to install the files at the correct place, be compatible with the currently installed libraries, etc. It's also easy to recompile from source.

The result of this script wouldn't have anything special or official, it's just files built using GitHub actions, that aren't guaranteed to work everywhere. Plus, it's just a subset of the architectures libsodium runs on. 

Since your use case seems to be .NET projects, there's already a [NuGET package for libsodium](https://www.nuget.org/packages/libsodium), that is kept up to date, and tries to match the requirements for the targeted .NET versions. It's used by NSec and other bindings, so it seems to work :)

js-9:
Looks to me like that nuget package doesn't contain android support, for example.
Anyways, it was an "in case you find this useful" thing, and it's easy enough to add other architectures and systems as needed.

jedisct1:
We should add Android builds to the NuGET package, then :)

PR #1342, opened 2024-01-09
feat:Supports identification of the HarmonyOS system

icoder20150719:
build for harmonyos platform 

https://device.harmonyos.com/en/


jedisct1:
That file is part of a completely distinct project.

The file includes an explanation on how to send patches, so that all projects using it can take advantage of them:

https://github.com/autotools-mirror/autoconf/blob/9ae8d7f61fee806dd078dfea5251a52019531c06/build-aux/config.sub#L30-L46

PR #1345, opened 2024-01-17
Move includes in crypto_onetimeauth_poly1305.h outside of extern C

fiedukow:
Includes being outside of `extern "C"` are de-facto standard in all other include files of libsodium.

At the same time, having inside the `extern "C"` is causing problem with C++ vs. C toolchain being confused about which libraries should be linked. This was especially painful while working on `swift-sodium` integration with windows.


jedisct1:
Thank you!

PR #1356, opened 2024-02-29
fixed inability to create xcframework in Symbolic Link

caobug:
```
Creating Clibsodium.xcframework...
error: cannot compute path of binary 'Path(str: "/private/tmp/libsodium-stable/libsodium-apple/macos/lib/libsodium.a")' relative to that of '/tmp/libsodium-stable/libsodium-apple/macos/lib/libsodium.a' ls: /tmp/libsodium-stable/libsodium-apple/Clibsodium.xcframework: No such file or directory
```
```
ls -l /tmp
lrwxr-xr-x@ 1 root  wheel  11 12 15 22:43 /tmp -> private/tmp
```


jedisct1:
Thank you!

PR #1364, opened 2024-04-03
chore: fix typos

xiaoxianBoy:
fix a typo: `passwore`->`password`


jedisct1:
Not a typo. This is the same hash as the previous test, with an intentionally wrong input. The purpose of the test is to check that verification doesn't pass, even with a small difference.

PR #1369, opened 2024-04-25
Support for `netX-maccatalyst` TFM in .NET

MartyIX:
This PR adds support for using libsodium in Mac Catalyst .NET projects.

## Testing

I was testing this PR on my `next` branch (https://github.com/MartyIX/libsodium/actions/runs/8745576446) and it created: [nuget-package](https://github.com/MartyIX/libsodium/actions/runs/8745576446/artifacts/1427739121)

The nuget package can be tested with my demo application on Mac Catalyst: 

https://github.com/MartyIX/MaccatalystLibsodium

The only modified file in the "empty Mac Catalyst" project is this file: https://github.com/MartyIX/MaccatalystLibsodium/blob/main/MaccatalystLibsodium/AppDelegate.cs where I added some calls into libsodium library.

### Screenshot

image
 

## Resources

* https://github.com/jedisct1/libsodium/discussions/1235#discussioncomment-4443263
* https://github.com/jedisct1/libsodium/pull/1238#issuecomment-1383111044

### DIFF
--- a/packaging/dotnet-core/libsodium.pkgproj
+++ b/packaging/dotnet-core/libsodium.pkgproj
@@ -34,7 +34,9 @@
     <Content Include="runtimes/linux-musl-arm/native/libsodium.so" PackagePath="runtimes/linux-musl-arm/native/" />
     <Content Include="runtimes/linux-musl-arm64/native/libsodium.so" PackagePath="runtimes/linux-musl-arm64/native/" />
     <Content Include="runtimes/osx-x64/native/libsodium.dylib" PackagePath="runtimes/osx-x64/native/" />
-    <Content Include="runtimes/osx-arm64/native/libsodium.dylib" PackagePath="runtimes/osx-arm64/native/" />    
+    <Content Include="runtimes/osx-arm64/native/libsodium.dylib" PackagePath="runtimes/osx-arm64/native/" />
+    <Content Include="runtimes/maccatalyst-x64/native/libsodium.a" PackagePath="runtimes/maccatalyst-x64/native/" />
+    <Content Include="runtimes/maccatalyst-arm64/native/libsodium.a" PackagePath="runtimes/maccatalyst-arm64/native/" />
   </ItemGroup>
 
 </Project>
--- a/packaging/dotnet-core/libsodium.pkgproj
+++ b/packaging/dotnet-core/libsodium.pkgproj
@@ -30,11 +30,11 @@
     <Content Include="runtimes/linux-x64/native/libsodium.so" PackagePath="runtimes/linux-x64/native/" />
     <Content Include="runtimes/linux-arm/native/libsodium.so" PackagePath="runtimes/linux-arm/native/" />
     <Content Include="runtimes/linux-arm64/native/libsodium.so" PackagePath="runtimes/linux-arm64/native/" />
-    <Content Include="runtimes/linux-musl-x64/native/libsodium.so" PackagePath="runtimes/linux-musl-x64/native/" />
-    <Content Include="runtimes/linux-musl-arm/native/libsodium.so" PackagePath="runtimes/linux-musl-arm/native/" />
-    <Content Include="runtimes/linux-musl-arm64/native/libsodium.so" PackagePath="runtimes/linux-musl-arm64/native/" />
-    <Content Include="runtimes/osx-x64/native/libsodium.dylib" PackagePath="runtimes/osx-x64/native/" />
-    <Content Include="runtimes/osx-arm64/native/libsodium.dylib" PackagePath="runtimes/osx-arm64/native/" />
+    <Content Include="runtimes/linux-musl-x64/native/libsodium.a" PackagePath="runtimes/linux-musl-x64/native/" />
+    <Content Include="runtimes/linux-musl-arm/native/libsodium.a" PackagePath="runtimes/linux-musl-arm/native/" />
+    <Content Include="runtimes/linux-musl-arm64/native/libsodium.a" PackagePath="runtimes/linux-musl-arm64/native/" />
+    <Content Include="runtimes/osx-x64/native/libsodium.a" PackagePath="runtimes/osx-x64/native/" />
+    <Content Include="runtimes/osx-arm64/native/libsodium.a" PackagePath="runtimes/osx-arm64/native/" />
     <Content Include="runtimes/maccatalyst-x64/native/libsodium.a" PackagePath="runtimes/maccatalyst-x64/native/" />
     <Content Include="runtimes/maccatalyst-arm64/native/libsodium.a" PackagePath="runtimes/maccatalyst-arm64/native/" />
   </ItemGroup>
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -118,9 +118,6 @@ jobs:
       - name: tests
         run: |
           cd zig-out/bin && ./run.sh
-      - name: debug-ls
-        run: |
-          ls -l -R zig-out/
       - uses: actions/upload-artifact@v3
         with:
           name: build-linux-musl-x64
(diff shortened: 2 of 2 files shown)


packaging/dotnet-core/libsodium.pkgproj
```diff
@@ -30,11 +30,13 @@
...
     <Content Include="runtimes/linux-arm/native/libsodium.so" PackagePath="runtimes/linux-arm/native/" />
     <Content Include="runtimes/linux-arm64/native/libsodium.so" PackagePath="runtimes/linux-arm64/native/" />
-    <Content Include="runtimes/linux-musl-x64/native/libsodium.so" PackagePath="runtimes/linux-musl-x64/native/" />
-    <Content Include="runtimes/linux-musl-arm/native/libsodium.so" PackagePath="runtimes/linux-musl-arm/native/" />
-    <Content Include="runtimes/linux-musl-arm64/native/libsodium.so" PackagePath="runtimes/linux-musl-arm64/native/" />
-    <Content Include="runtimes/osx-x64/native/libsodium.dylib" PackagePath="runtimes/osx-x64/native/" />
-    <Content Include="runtimes/osx-arm64/native/libsodium.dylib" PackagePath="runtimes/osx-arm64/native/" />    
+    <Content Include="runtimes/linux-musl-x64/native/libsodium.a" PackagePath="runtimes/linux-musl-x64/native/" />
+    <Content Include="runtimes/linux-musl-arm/native/libsodium.a" PackagePath="runtimes/linux-musl-arm/native/" />
+    <Content Include="runtimes/linux-musl-arm64/native/libsodium.a" PackagePath="runtimes/linux-musl-arm64/native/" />
```
ektrah, line 35: Is it intended that the precompiled binary for musl Linux is now a `.a`? As far as I know, this has to be a `.so`.

ektrah (commented):
@MartyIX Thanks, this looks like it's going in the right direction. 👍 

I guess I have to do something similar to this in the binding? https://github.com/ektrah/nsec/compare/master...ios

ektrah:
@jedisct1 BTW, it's possible to make prereleases of NuGet packages by setting the `<Version>` property in `libsodium.pkgproj` like this:

```diff
-<Version>1.0.21.0</Version>
+<Version>1.0.21.0-preview.1</Version>
```

packaging/dotnet-core/libsodium.pkgproj
alexrp: This is definitely wrong. .NET only uses static linking for a subset of Apple platforms (iOS etc). These should be `.so`.

(I would note that there's a high likelihood that these musl binaries are unusable anyway due to https://github.com/ziglang/zig/issues/11909, but that's a separate matter.)

alexrp: I'm unfortunately not familiar enough with libsodium's `build.zig` to know why this would be happening. A cursory glance at it seems to suggest that it *should* be building shared libraries by default.

alexrp: `linux-` RIDs should be `.so`, and `osx-` RIDs should be `.dylib`.

I'm not super familiar with Mac Catalyst, so take this with a grain of salt, but it *sounds* right that it would be `.a`; iOS has no concept of shared libraries, so I would assume the same is true of Mac Catalyst.

ektrah:
Version 1.0.19.1 is broken on Alpine Linux because the NuGet package does not contain a `runtimes/linux-musl-x64/native/libsodium.so`.

Package contents: https://nuget.info/packages/libsodium/1.0.19.1

Failing test: https://github.com/ektrah/nsec/actions/runs/8869195585/job/24349607561

jedisct1:
`runtimes/linux-musl-x64/native/libsodium.a` should be there.

Is a shared library required even for musl?

alexrp:
`linux-musl` in .NET refers to systems that use musl as a system-wide shared library, e.g. Alpine.

jedisct1:
I don't get how it used to work before, if it did. Zig can only create static libraries when using `musl`.

ektrah:
Assuming I picked the right commit, it looks like the `runtimes/linux-musl-x64/native/libsodium.so`  in 1.0.19 was built in an Alpine 3.13 container: https://github.com/jedisct1/libsodium/blob/b5bd5b877116948b7d68bfd3f308bdc2953cca6c/.github/workflows/dotnet-core.yml#L101-L124

alexrp:
The .NET SDK will currently only use static libraries under `runtimes/...` for certain Apple platforms (iOS and friends). Shipping a static library for a Linux-based platform under `runtimes/...` has no effect.

It actually seems like https://github.com/ziglang/zig/issues/11909 does *not* affect libraries:

```console
❯ zig cc test.c -target x86_64-linux-musl -shared -fPIC -o libtest.so^C
❯ bat test.c
       File: test.c
   1   #include <stdio.h>
   2   
   3   void hello_world()
   4   {
   5       printf("hello world\n");
   6   }
❯ zig cc test.c -target x86_64-linux-musl -shared -fPIC -o libtest.so
❯ file libtest.so
libtest.so: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
❯ readelf -d libtest.so

Dynamic section at offset 0x3c0 contains 16 entries:
  Tag        Type                         Name/Value
 0x0000000000000001 (NEEDED)             Shared library: [libc.so]
 0x000000000000001e (FLAGS)              BIND_NOW
 0x000000006ffffffb (FLAGS_1)            Flags: NOW
 0x0000000000000017 (JMPREL)             0x2f8
 0x0000000000000002 (PLTRELSZ)           24 (bytes)
 0x0000000000000003 (PLTGOT)             0x24c0
 0x0000000000000014 (PLTREL)             RELA
 0x0000000000000006 (SYMTAB)             0x200
 0x000000000000000b (SYMENT)             24 (bytes)
 0x0000000000000005 (STRTAB)             0x2d0
 0x000000000000000a (STRSZ)              40 (bytes)
 0x000000006ffffef5 (GNU_HASH)           0x278
 0x0000000000000004 (HASH)               0x2a0
 0x000000000000000c (INIT)               0x1394
 0x000000000000000d (FINI)               0x1397
 0x0000000000000000 (NULL)               0x0
```

(And so on for other architectures.)

So I would suggest just going back to shipping shared objects for musl as before.

jedisct1:
Ok, I think e77588fe8fae5d9948a9534aa39f32e1ba076407 is the key.

ektrah:
BTW, CentOS Linux 7 will reach end of life (EOL) on June 30, 2024.

.NET 8 has already dropped support for it; see https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md

.NET 7 and .NET 6 will reach end of life on May 14, 2024 and November 12, 2024.  
https://github.com/dotnet/core/blob/main/release-notes/7.0/supported-os.md  
https://github.com/dotnet/core/blob/main/release-notes/6.0/supported-os.md

From my point of view I think it's okay to drop support for those ancient Linux versions in the libsodium NuGet package now and go forward with the platforms supported by .NET 8.

jedisct1:
Oh, this is good news!

ektrah:
Will do. I think that's not the only change needed for iOS/tvOS/MacCatalyst support though, is it?

Try this: Add both NuGet packages to your project, i.e. both libsodium-core 1.3.5 and libsodium 1.0.19.2. This should cause libsodium-core 1.3.5 to use libsodium 1.0.19.2 instead of 1.0.19. I don't think this will fully work yet, though, as the `__Internal` magic is still missing in libsodium-core 1.3.5.

ektrah:
Done. [v1.4.0-preview.1](https://github.com/ektrah/libsodium-core/releases/tag/v1.4.0-preview.1)

Please try it and provide feedback [here](https://github.com/ektrah/libsodium-core/issues/80).

PR #1405, opened 2024-09-09
Android-build, set NDK_API_VERSION using version from ANDROID_NDK_HOME

jdelromano:
I got stuck in generating the libraries because of a NDK version mismatch. I made this change and it worked.

With this change, the value of NDK_API_VERSION is taken from the file "${ANDROID_NDK_HOME}/source.properties".
NDK_PLATFORM will then be set accordingly.

### DIFF
--- a/dist-build/android-build.sh
+++ b/dist-build/android-build.sh
@@ -1,17 +1,19 @@
 #! /bin/sh
 
-if [ -z "$NDK_PLATFORM" ]; then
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi,

I'd like to avoid breaking changes and fragile parsing if possible. 

Why not just set `NDK_PLATFORM` to the desired target? 

Guidance can be printed if it is not found given the installed NDK.

PR #1407, opened 2024-09-11
Add support for Android's new 16K page sizes

bidrohi:


meiyese:
Hi, @bidrohi @jedisct1 

Do we really need to do this for ARMv7-A and x86? Based on the [documentation](https://developer.android.com/guide/practices/page-sizes#alignment-use-script), it looks like we don't have to.

bidrohi:
Technically we don't because Android is also moving forward with 64-bit only, but I added it there for consistency.

PR #1415, opened 2024-10-02
blake2b: Add aarch64 NEON acceleration

r1mikey:
Import NEON acceleration from blake2b-ref. Copyrights reflect those in the upstream files, specifically:
  https://github.com/BLAKE2/BLAKE2/blob/master/neon/blake2b-neon.c

I've restricted this to aarch64 as I do not have an environment in which to test armv7-a support.

Successfully built and run on both aarch64 (Nvidia Grace) and x86_64. Tested on both with `make verify`.

### DIFF
--- a/src/libsodium/Makefile.am
+++ b/src/libsodium/Makefile.am
@@ -35,6 +35,9 @@ libsodium_la_SOURCES = \
 	crypto_generichash/blake2b/generichash_blake2.c \
 	crypto_generichash/blake2b/ref/blake2.h \
 	crypto_generichash/blake2b/ref/blake2b-compress-ref.c \
+	crypto_generichash/blake2b/ref/blake2b-compress-neon.c \
+	crypto_generichash/blake2b/ref/blake2b-compress-neon.h \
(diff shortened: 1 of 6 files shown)


r1mikey:
Let's hold on merging this while I do some benchmarking as well.

jedisct1:
Something like this would be required after including `<arm_neon.h>`:

```c
#    ifdef __clang__
#        pragma clang attribute push(__attribute__((target("neon"))), apply_to = function)
#    elif defined(__GNUC__)
#        pragma GCC target("+simd")
#    endif
```

r1mikey:
This is not worth pursuing for 64 bit Arm systems based on my benchmarks. Performance of the NEON implementation comes in at around 66% of the C implementation. There are a few reasons for this:
* Instruction density is 89% of the C implementation (good)
* Instructions per cycle is at 46% of the C implementation (very bad)
* 54.6% of slots are backend bound (vs 21.3% for the C implementation - also very bad)
* The benchmark completes in 10,718.84 msec compared to 5,542.82 msec for the C implementation

jedisct1:
Thanks for having taken the time to run these benchmarks!

PR #1455, opened 2025-05-08
Add Android binaries to NuGet package

jesuslpm:
Runtime identifiers added: `android-arm`, `android-arm64`, `android-x64`, `android-x86`


jedisct1:
Looks good!

Thank you Jesus!

PR #1470, opened 2025-07-01
Fix block alignment calculation in Poly1305 update

harley005:
The previous implementation used an incorrect formula for calculating the padding length during the Poly1305 update. The calculation `(0x10 - (sizeof block) + mlen) & 0xf` was incorrect and could lead to misaligned input blocks.


jedisct1:
This is mentioned in the comment on the following line :) 

It doesn't have any security implications. However, changing this would be a breaking change, making all the previous cipher texts impossible to decrypt. So, this is not going to change.

PR #1473, opened 2025-08-20
Add aarch64 scalarmult_curve25519 implementation

Nicoshev:
Adding AWS aarch64 scalarmult_curve25519 implementation, published here: https://github.com/awslabs/s2n-bignum/tree/main/arm/curve25519

It provides almost double throughput, when compared to the ref algorithm.

### DIFF
--- a/src/libsodium/Makefile.am
+++ b/src/libsodium/Makefile.am
@@ -71,6 +71,11 @@ libsodium_la_SOURCES = \
 	crypto_pwhash/argon2/pwhash_argon2id.c \
 	crypto_pwhash/crypto_pwhash.c \
 	crypto_scalarmult/crypto_scalarmult.c \
+	crypto_scalarmult/curve25519/aarch64/curve25519_x25519.c \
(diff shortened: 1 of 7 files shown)


jedisct1:
Hi,

Thanks for sharing this.

It’s a lot of code with its own style, and it won’t be easy to maintain. I’m happy to take it on under a support contract, but I can’t do it in my spare time.

It also reimplements Edwards25519 arithmetic and includes the large Edwards25519 base point precomputation table. That’s a lot of duplication, especially since it will only be used for X25519.

While it may solve a specific problem you have, I don’t think this is the right way to improve performance on AArch64.
Something that only does variable-base multiplication, like sandy2x, would be more reasonable. There’s an existing NEON implementation written in qhasm that may be easier to integrate and maintain.

In the long run, optimized field operations from these implementations could replace the current C code, so all Edwards25519 operations benefit, and, as a side effect, fixed-base X25519 would benefit too (the projection itself is very fast).

PR #1474, opened 2025-08-27
make build.zig emit position independent code by default

stef:
i noticed that build.zig doesn't pass `-fPIC` as a compile flag, so i made it and option and default.

one thing, due to the iomageddon in zig 0.15.1 i made this change against the git revision before compatibility with zig 0.15.1, for i fear a bunch of projects will have trouble for some time to update their code to zig 0.15.1 (myself definitely) and for those (and most importantly myself) i'd appreciate if there would be a branch that has 0.14.1 compatibility and this patch. if i'm not imposing with this too much.

with this patch i'm able to cross-compile https://github.com/stef/klutshnik as pie-static binary. which pleases me very much!

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -154,6 +154,7 @@ pub fn build(b: *std.Build) !void {
     const benchmarks_iterations = b.option(u32, "iterations", "Number of iterations for benchmarks.") orelse 200;
     var build_static = b.option(bool, "static", "Build libsodium as a static library.") orelse true;
     var build_shared = b.option(bool, "shared", "Build libsodium as a shared library.") orelse true;
+    const build_pie = b.option(bool, "pie", "Build libsodium as position-independent code.") orelse true;
 
     const build_tests = b.option(bool, "test", "Build the tests (implies -Dstatic=true)") orelse true;
 
@@ -221,6 +222,7 @@ pub fn build(b: *std.Build) !void {
             "-fno-strict-aliasing",
             "-fno-strict-overflow",
             "-fwrapv",
+            if (build_pie) "-fPIC" else "-fno-PIC",
             "-flax-vector-conversions",
         };
 


stef:
upon reconsideration, i don't really need a dedicated branch in your fine repo. i can just use my own:

```
zig fetch --save=libsodium 'git+https://github.com/stef/libsodium.git#2ab5db6ec8d7a9c15b08329006e09c44dd51343e'
```
works perfectly fine!

jedisct1:
`PIE` or `PIC`? They are different flags.

In the Autoconf script we enable both unconditionally, except on platforms with known issues.

Instead of adding a new option, maybe we could do the same here. Would that break anything, for example, when integrating with Zig?

jedisct1:
BTW, Klutshnik looks really cool!

stef:
> `PIE` or `PIC`? They are different flags.

i propose a new flag: -facepalm. :)
indeed, PIC for the libs, PIE for the executables. my bad. also, TIL. so yes, for libsodium i really mean PIC.

> In the Autoconf script we enable both unconditionally, except on platforms with known issues.
> 
> Instead of adding a new option, maybe we could do the same here. Would that break anything, for example, when integrating with Zig?

you propose to not have any option, just enable pic by default? worksforme!

> BTW, Klutshnik looks really cool!

thank you! i think you might be delighted to learn about this branch then: https://github.com/stef/libsodium/tree/zephyr-port - which is an improvement on  https://github.com/pmessan/libsodium-zephyr - which in turn enables https://github.com/stef/klutshnik-zephyr - so you can run your klutshnik servers via USB or BLE on microcontrollers...

stef:
i just removed all this misnamed options stuff, and just made `-fPIC` default. hope this is what you wanted. cheers, and thank you so much for your fine product! kudos,
s

PR #1487, opened 2025-11-17
Support for Visual Studio 2026

neonbyte1:
I've duplicated the `builds/msvc/vs2022` directory, renamed it into `vs2026` and successfully retargeted the project to the latest platform toolset (v145). I was also able to build every configuration for every possible target without any warnings or errors.

### DIFF
--- a/builds/msvc/vs2026/libsodium.import.props
+++ b/builds/msvc/vs2026/libsodium.import.props
@@ -1,0 +1,52 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup Label="Globals">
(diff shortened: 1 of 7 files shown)


neonbyte1:
Not sure if I should add the files to `EXTRA_DIST ` in  `builds/Makefile.am`.

jedisct1:
Hi,

Thanks!

The project files in `builds/msvc` are all automatically generated by the `regen-msvc.py` script, so they don't need to be copied or modified manually.

Support for Visual Studio was added to the `stable` branch here:  6093bc954ae50f6989f550d5651fb200db7db806

Unfortunately, Azure Pipelines runners don't have MSVC 2026 installed yet, so the binary packages don't include builds using that compiler yet. They should appear as soon as Microsoft adds that version to their environment.

neonbyte1:
Ah I didn't know that these were generated using a script. I saw your changes this morning so I guess this PR won't be needed anymore 🙈

PR #1505, opened 2026-01-08
build.zig: fix using as a dependency in other projects

torque:
This was originally fixed in 5191b40accb6c860c6d1292c4149c7bea42aca63 but regressed in a8cc2d1f7228252e2d1150cd04dd234b6d52df53.


jedisct1:
Thank you!

PR #1515, opened 2026-02-20
Feat/ristretto255 h2c tests

wmcelderry:
Adding tests for ristretto255_h2c  given as part of test vectors for RFC9497 (OPRFs)

### DIFF
--- a/test/default/Makefile.am
+++ b/test/default/Makefile.am
@@ -30,6 +30,7 @@ EXTRA_DIST = \
 	codecs.exp \
 	core_ed25519.exp \
 	core_ed25519_h2c.exp \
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks for putting this together and sharing it. I had a look, and it feels more like a work-in-progress draft at this stage. 

There are a few inconsistencies with the existing code style, some commented-out sections, TODOs, and debugging notes that would need to be cleaned up.

Tests should also be focused on the primitive they are testing, not implement unrelated protocols, because tests are also used for benchmarking.

It’s totally fine for experimentation or testing, but I don’t think it’s ready to be merged.

PR #1517, opened 2026-03-06
test: add wycheproof tests

tob-scott-a:
This includes a Python script to fetch the latest test vectors from https://github.com/C2SP/wycheproof and update the header files in test/wycheproof/vectors, as well as a dedicated GitHub Actions workflow to run these tests in CI.

### DIFF
--- a/.github/workflows/wycheproof.yml
+++ b/.github/workflows/wycheproof.yml
@@ -8,11 +8,16 @@ on:
 
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   wycheproof:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
 
       - name: Install build dependencies
         run: |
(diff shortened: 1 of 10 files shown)


jedisct1:
Given what happened after Wycheproof's author got fired from Google, I'm wondering if this is not an opportunity to adopt [Rooterberg](https://github.com/bleichenbacher-daniel/Rooterberg) instead.

As a bonus, it has test vectors for AEGIS with 256-bit tags, and it generally much cleaner and well organized than Wycheproof. This is also an opportunity to give feedback and contribute to it.

tob-scott-a:
I'm not sure what you're referring to, but Wycheproof is now a [C2SP](https://github.com/C2SP/wycheproof) project, not a Google one.

jedisct1:
The repository was eventually transferred, but as far as I know, Daniel, who created and maintained the project, lost access to it in the process.

Code from the original repository was deleted, and he never regained access to the tooling he had written to generate the test vectors.

He tentatively reached out to various parties involved, but received very few responses to his questions and was effectively prevented from continuing to contribute to his own project.

Setting aside the surrounding controversy, and with the hope that both projects will eventually merge, Rooterberg is technically cleaner, and IMHO a better fit for libsodium.

PR #1518, opened 2026-03-06
ci: use zizmor to prevent supply chain attacks via GHA

tob-scott-a:


jedisct1:
GitHub actions don't publish anything, so the impact of a supply chain attack would be quite limited.

The `dotnet-core.yml` workflow builds .NET packages, though, so pinning the actions hashes can be a good thing, thanks!

PR #1520, opened 2026-03-07
Zero AEGIS cipher state and sensitive locals after use

cpsource:
## Summary

- Add `sodium_memzero()` calls to wipe key-derived cipher state, computed MAC, and scratch buffers (`src`, `dst`) after use in AEGIS-128L and AEGIS-256 `encrypt_detached` and `decrypt_detached` functions.
- Replace plain `memset(m, 0, mlen)` with `sodium_memzero(m, mlen)` when clearing plaintext on authentication failure, preventing the compiler from optimizing the wipe away under `-O3`.
- Aligns AEGIS with the existing practice in `aes256gcm` and `chacha20poly1305`, which already zero sensitive state after use.

## Files changed

- `src/libsodium/crypto_aead/aegis128l/aegis128l_common.h`
- `src/libsodium/crypto_aead/aegis256/aegis256_common.h`

## Test plan

- [x] `make check TESTS="aead_aegis128l aead_aegis256"` passes with no regressions
- [x] Verified all 46 AEGIS-128L and 52 AEGIS-256 KAT vectors still pass
- [x] Confirmed `sodium_memzero` is available via `utils.h` (already included by all backends: aesni, armcrypto, soft)

### DIFF
--- a/src/libsodium/crypto_aead/aegis128l/aegis128l_common.h
+++ b/src/libsodium/crypto_aead/aegis128l/aegis128l_common.h
@@ -164,6 +164,7 @@ encrypt_detached(uint8_t *c, uint8_t *mac, size_t maclen, const uint8_t *m, size
     CRYPTO_ALIGN(RATE) uint8_t src[RATE];
     CRYPTO_ALIGN(RATE) uint8_t dst[RATE];
     size_t                     i;
+    int                        ret;
 
     aegis128l_init(k, npub, state);
 
@@ -188,7 +189,13 @@ encrypt_detached(uint8_t *c, uint8_t *mac, size_t maclen, const uint8_t *m, size
         memcpy(c + i, dst, mlen % RATE);
     }
 
-    return aegis128l_mac(mac, maclen, adlen, mlen, state);
+    ret = aegis128l_mac(mac, maclen, adlen, mlen, state);
+
+    sodium_memzero(state, sizeof state);
+    sodium_memzero(src, sizeof src);
+    sodium_memzero(dst, sizeof dst);
+
+    return ret;
 }
 
 static int
@@ -242,9 +249,13 @@ decrypt_detached(uint8_t *m, const uint8_t *c, size_t clen, const uint8_t *mac, 
             ret = crypto_verify_32(computed_mac, mac);
         }
     }
+    sodium_memzero(state, sizeof state);
+    sodium_memzero(src, sizeof src);
+    sodium_memzero(dst, sizeof dst);
+    sodium_memzero(computed_mac, sizeof computed_mac);
     if (ret != 0) {
         if (m != NULL) {
-            memset(m, 0, mlen);
+            sodium_memzero(m, mlen);
         }
         return ret;
     }
--- a/src/libsodium/crypto_aead/aegis256/aegis256_common.h
+++ b/src/libsodium/crypto_aead/aegis256/aegis256_common.h
@@ -147,6 +147,7 @@ encrypt_detached(uint8_t *c, uint8_t *mac, size_t maclen, const uint8_t *m, size
     CRYPTO_ALIGN(RATE) uint8_t src[RATE];
     CRYPTO_ALIGN(RATE) uint8_t dst[RATE];
     size_t                     i;
+    int                        ret;
 
     aegis256_init(k, npub, state);
 
@@ -171,7 +172,13 @@ encrypt_detached(uint8_t *c, uint8_t *mac, size_t maclen, const uint8_t *m, size
         memcpy(c + i, dst, mlen % RATE);
     }
 
-    return aegis256_mac(mac, maclen, adlen, mlen, state);
+    ret = aegis256_mac(mac, maclen, adlen, mlen, state);
+
+    sodium_memzero(state, sizeof state);
+    sodium_memzero(src, sizeof src);
+    sodium_memzero(dst, sizeof dst);
+
+    return ret;
 }
 
 static int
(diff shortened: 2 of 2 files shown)


jedisct1:
The state is not invertible, so this is less of a requirement for AEGIS. It will leak the MAC. Not sure that this is an issue in a nonce-respecting setting. 

Zeroing the ciphertext with sodium_memzero could also be way too slow for some applications.

So we should be a little bit more selective here.

jedisct1:
That really feels AI-generated.

cpsource:
The first pass of the code review was AI, but I gave it a second pass
before I created the PR.

The main point of the PR was: Why put a memset/return when -O3 removes the
memset? You code one thing and think it's done when it isn't.

 If sodium_memzero is too slow, which it might be:

(again from AI, sorry ... well not sorry)

// Option 5: fast + safe
static inline void secure_zero(void *p, size_t n) {
    memset(p, 0, n);
    __asm__ __volatile__("" : : "r"(p) : "memory");
}

The "r"(p) tells gcc that the asm "uses" the pointer, and "memory" says it
might read/write any memory. So gcc must keep the memset alive because the
asm might depend on the buffer being zeroed. But the asm itself emits
nothing — zero overhead.

That's the one I'd use. One line, no runtime cost, gcc can still vectorize
the memset itself, and -O3 won't touch it

Cal

cpsource:
And, while we are at it, here's a couple of other things you should know
about:

 1. [LOW] Bitwise & instead of logical && — aead_aegis128l.c:155 and
aead_aegis256.c:154

  if (sodium_runtime_has_aesni() & sodium_runtime_has_avx()) {

  Should be &&. Both functions return int (0 or 1), so the result is the
same here, but & defeats short-circuit evaluation
   — has_avx() is always called even if has_aesni() returns 0. It's
functionally harmless (tiny perf cost at init time, no
   correctness issue), but it's not idiomatic and could confuse code
reviewers. This may be intentional to avoid branch
  prediction side effects, though.

  2. [LOW] No empty-message (mlen=0) test vector — in both
test/default/aead_aegis128l.c and test/default/aead_aegis256.c

  The MAC-only path (encrypt with empty plaintext, AD-only authentication)
is valid but has no KAT vector confirming the
  MAC output is correct.

  Neither of these is a security vulnerability — one is a style nit, the
other is a test coverage gap. Up to you whether
  they're worth reporting. The bitwise & one might be intentional given the
codebase style, so it could be worth asking
  the maintainer rather than filing a fix.

jedisct1:
> The main point of the PR was: Why put a memset/return when -O3 removes the memset?

AEGIS is not RUP secure, meaning that applications should not disclose unverified plaintext.

That's the purpose of the `memset()`, as an extra layer of protection in case an application decrypts a ciphertext and  doesn't check if decryption was successful in spite of the `warn_unused_result` attribute.

The `-O3` optimization level has nothing to do with that. `memset()` may be elided if the output is not used.  Not if it is actually used :) 

` __asm__ __volatile__("" : : "r"(p) : "memory");` is one of the paths `sodium_memzero` already takes, but inline assembly is not supported by all compilers, especially with this syntax.

Against cold-boot attacks, even though this is mostly theatre since all the interesting bits are in SIMD registers, not on the stack,  there are a few things we can zero, but certainly not what the LLM suggested.

> `if (sodium_runtime_has_aesni() & sodium_runtime_has_avx()) {`
> Should be &&. Both functions return int (0 or 1).

`&` is perfectly fine.

jedisct1:
Here's an illustration that zeroing vector registers is not only useless, but actually harmful:

https://godbolt.org/z/G37bch76P

Notice that in `zok2`, `zmm1` is not zeroed, even though your LLM may pretend it is. The actual difference is that in addition to what `zok1` does, it now allocates unnecessary space on the stack, only to write zeroes there.

It's not zeroing any sensitive data (it's still in `zmm1`), it's actually not zeroing any active data, it's just wasting code, CPU cycles and stack usage for a placebo operation.

`zok3` doesn't call a generic zeroing function, and tries to zero the register directly. As you can see, that doesn't work either, `zmm1` is untouched.

We could zero it in the inline assembly code, but it's unlikely to work either, since the compiler can copy the value into another register, such as reusing `zmm0`, leaving `zmm1` still untouched.

The only reliable way to do zero secrets is to write everything in assembly.

A still unreliable but maybe okay alternative is to have a function that wipes the entire stack, and all the registers.
`sodium_stackzero` does the former, it may eventually be extended to also do the latter. 

But this is an expensive operation, so it's something that application can do after they are done with a sequence of operations involving a set of secrets. Not something we can realistically do in every single function call.

cpsource:
Frank,
  Excellent points!

Re: assembly code ???

I took your challenge to move everything to assembler and added the
*vzeroall* instruction to aegis128l_aesni.s. I recompiled libsodium and
reran the QA tests, and all passed. (The code required the patch that I
sent you, ie // Option 5: fast + safe
static inline void secure_zero(void *p, size_t n) {
    memset(p, 0, n);
    __asm__ __volatile__("" : : "r"(p) : "memory");
}
I did not 'kit' this solution nor extend it to all your round loops. If you
want me to do this, I can submit a PR. Let me know.

I've attached the REAME-vzeroall.md detailing this work along with the
python editing program.

And running under Linux/ubuntu raises other concerns. See attached. And
don't get me started on EMI/RFI concerns from the PC this is run on. You
can, these days, point an antenna at the PC from some blocks away and watch
what the PC does.

I haven't evaluated LTO:

And there's a second issue the developer didn't mention but that's been
discussed in the OpenSSL community: **link-time optimization (LTO)**. With
LTO enabled, the compiler can see across translation units. A sufficiently
aggressive LTO pass could theoretically determine that `memset_func` is
initialized to `memset`, never modified (the `volatile` qualifier makes
this less likely but the spec is murky on whether the compiler could prove
no other thread modifies it), and optimize accordingly. In practice current
compilers don't do this, but it's not guaranteed by the standard — it works
by relying on compiler behavior rather than language guarantees.
                               That's why C11 added `memset_s` and POSIX
added `explicit_bzero` — to provide an actual language/OS level guarantee
rather than depending on tricking the optimizer.

Cal

jedisct1:
The libsodium documentation explicitly says that LTO should not be used. https://doc.libsodium.org/doc/installation

jedisct1:
There are implementations of AEGIS in assembly: https://github.com/aegis-aead/aegis-jasmin

Jasmin ensures that secrets are cleared from memory, including registers.

PR #1548, opened 2026-07-08
configure: require a declaration for getpid

omartijn:
AC_CHECK_FUNCS' link test false-passes when targeting the MSVC CRT: getpid() resolves through the oldnames.lib compatibility alias for _getpid(), but no header declares it, so compilation of randombytes_internal_random.c later fails under C99+ implicit-declaration rules. Chain the function check behind AC_CHECK_DECL so HAVE_GETPID is only defined when a declaration exists, matching the MSVC project files, which never define it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -979,7 +979,11 @@ if (&getentropy != NULL) {
 ])
 
 AS_IF([test "x$WASI" = "x"],[
-  AC_CHECK_FUNCS([getpid])
+  dnl A plain link check false-passes when targeting the MSVC CRT: getpid()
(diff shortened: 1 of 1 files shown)


jedisct1:
I don't understand. 

The library comes with Visual Studio project files to be compiled with MSVC.

Alternatively, the zig build system can also be used. 

autoconf scripts are not for MSVC. At least they were never tested or expected to ever work with MSVC and looks like no one ever tried that 😆

jedisct1:
AI generated slop is also not accepted in this project.

omartijn:
Ah lovely - "I don't understand so I'll just close it".

PR #1552, opened 2026-08-20
test: include <signal.h> unconditionally

lzcunt:
HAVE_CATCHABLE_SEGV comes from an AC_RUN_IFELSE probe, so it is never defined when cross compiling. There is a usage of `signal` that is not gated by this check.

<signal.h> is part of the C standard library and always available, so just include it.


jedisct1:
If `signal.h` isn't included, signal names shouldn't be defined, so `signal()` calls should not be compiled in.

What exact environment, compiler and target are you using where this is a problem?

jedisct1:
Your PR is a revert of a8fa837a - And that would break compilation to webassembly (nope: `signal.h` isn't always available).

It would also silently weaken the tests.  With these changes, `signal(SIGABRT, segv_handler` compiles in wherever `HAVE_CATCHABLE_SEGV` is undefined,  including ASan builds. I let you find why this is a regression :)

jedisct1:
Tests are meant to run on the host the library was compiled on. This is what the `make check` command does. Running them independently when using Autoconf is undocumented, and TBH, I have no idea how to do that :)

For cross-compilation, and to produce standalone test executables, libsodium should be compiled with Zig.

And `signal.h` doesn't seem to be an issue on any of the supported Zig targets.

So, sorry, but I have to close this, as it would break things for existing users.

lzcunt:
> If `signal.h` isn't included, signal names shouldn't be defined, so `signal()` calls should not be compiled in.

The environment I worked on apparently leaked those signal names from sys/types.h

> What exact environment, compiler and target are you using where this is a problem?

x86_64-linux-unknown-mlibc target, x86_64-linux-unknown-gnu host, gcc

lzcunt:
> Tests are meant to run on the host the library was compiled on. This is what the `make check` command does. Running them independently when using Autoconf is undocumented, and TBH, I have no idea how to do that :)

Nix runs package tests when cross-compiling if binaries of the target platform can be executed on the build platform (e.g. same architecture same OS cross-libc or binfmt emulation of an entirely different target). I encountered this when I was cross-compiling with nix.

> For cross-compilation, and to produce standalone test executables, libsodium should be compiled with Zig.
> 
> And `signal.h` doesn't seem to be an issue on any of the supported Zig targets.
> 
> So, sorry, but I have to close this, as it would break things for existing users.
> 

Understandable, thanks for your time anyway

